PR #6, opened 2021-11-16
PLUGIN-980: Display Output Schema

sgarg-CS:
[PLUGIN-980](https://cdap.atlassian.net/browse/PLUGIN-980)
Show schema on click of Validate button for Service Now Batch Source Plugin

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSource.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSource.java
@@ -35,6 +35,7 @@ import io.cdap.plugin.common.SourceInputFormatProvider;
 import io.cdap.plugin.servicenow.source.util.ServiceNowConstants;
 import io.cdap.plugin.servicenow.source.util.ServiceNowTableInfo;
 import io.cdap.plugin.servicenow.source.util.SourceQueryMode;
+
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
 import org.slf4j.Logger;
@@ -74,6 +75,10 @@ public class ServiceNowSource extends BatchSource<NullWritable, StructuredRecord
     // Since we have validated all the properties, throw an exception if there are any errors in the collector.
     // This is to avoid adding same validation errors again in getSchema method call
     collector.getOrThrowException();
+    if (conf.shouldGetSchema()) {
+      stageConfigurer.setOutputSchema(
+        ServiceNowInputFormat.fetchTableInfo(conf.getQueryMode(collector), conf).get(0).getSchema());
+    }
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSource.java
```diff
@@ -74,6 +74,8 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     // Since we have validated all the properties, throw an exception if there are any errors in the collector.
     // This is to avoid adding same validation errors again in getSchema method call
     collector.getOrThrowException();
+    stageConfigurer.setOutputSchema(
+      ServiceNowInputFormat.fetchTableInfo(conf.getQueryMode(collector), conf).get(0).getSchema());
```
rmstar, line 78: Please check if this works with macros enabled. What happens if query mode, table name are macros?

sgarg-CS: hi @rmstar  

No, the schema won't be generated if the query mode and table name are set as Macro.

If we set these fields as Macro, then the following error appears at the time of validation.
**Unsupported query mode value: null Supported modes are: Reporting, Table**

Validation Error when Query Mode and Table Name are set as Macro

rmstar: Yes, so you need to handle this case. We should skip validation in configurePipeline for fields that have macros

sgarg-CS: The logic to skip validation in `configurePipeline` for fields that have macros already exists in the codebase. I've added the shouldGetSchema() method to check each mandatory field that can have Macro before setting the output schema. 171aa6eeb43b3413335edf8505017cd5ede924ad
This will prevent the above error.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
```diff
@@ -452,4 +453,14 @@ public boolean shouldConnect() {
       !containsMacro(ServiceNowConstants.PROPERTY_USER) &&
       !containsMacro(ServiceNowConstants.PROPERTY_PASSWORD);
   }
+
+  protected boolean shouldGetSchema() {
```
rmstar, line 457: Why is this protected? Can we make it public?

sgarg-CS: At the time of writing this method, I was referring to a similar method in a different connector that had `protected` as its access modifier.

Have changed it to `public`
8fbc7cfe28c54cafac8ef18493777c7657cb47e5

rmstar (approved):
please squash commits

sgarg-CS:
Commits have been squashed.

PR #7, opened 2021-11-17
PLUGIN-979: Remove Table Name Field

sgarg-CS:
[PLUGIN-979](https://cdap.atlassian.net/browse/PLUGIN-979)
Remove Table Name Field from ServiceNow Batch Source Plugin configuration

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
@@ -115,18 +115,18 @@ public class ServiceNowSourceConfig extends PluginConfig {
   /**
    * Constructor for ServiceNowSourceConfig object.
    *
-   * @param referenceName   The reference name
-   * @param queryMode       The query mode
+   * @param referenceName The reference name
+   * @param queryMode The query mode
    * @param applicationName The application name
-   * @param tableName       The table name
-   * @param clientId        The Client Id for ServiceNow
-   * @param clientSecret    The Client Secret for ServiceNow
+   * @param tableName The table name
+   * @param clientId The Client Id for ServiceNow
+   * @param clientSecret The Client Secret for ServiceNow
    * @param restApiEndpoint The rest API endpoint for ServiceNow
-   * @param user            The user id for ServiceNow
-   * @param password        The password for ServiceNow
-   * @param valueType       The value type
-   * @param startDate       The start date
-   * @param endDate         The end date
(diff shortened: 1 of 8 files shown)


rmstar:
Can you clarify why we don't need to set the table name in the schema when running in reporting mode?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
```diff
@@ -173,7 +163,7 @@ public SourceQueryMode getQueryMode(FailureCollector collector) {
     }
 
     collector.addFailure("Unsupported query mode value: " + queryMode,
-      String.format("Supported modes are: %s", SourceQueryMode.getSupportedModes()))
+        String.format("Supported modes are: %s", SourceQueryMode.getSupportedModes()))
```
rmstar, line 166: remove added whitespaces/indent changes here and elsewhere

sgarg-CS: the extra whitespaces/indent changes have been removed. 9753ec9e11296d5d686020e54a7f2716cf18eb04

sgarg-CS:
> Can you clarify why we don't need to set the table name in the schema when running in reporting mode?

In the reporting mode, the plugin fetches the list of all the ServiceNow's data tables for the application name provided in plugin configuration(e.g. Contract Management, Product Catalog, Procurement) and their details.

In the source code, the following logic is used to get the table names and the metadata for each table.

    // When mode = Reporting, get the list of tables for application name provided in plugin config
    // and then fetch details from each of the tables.
    List<ServiceNowTableInfo> tableInfos = new ArrayList<>();

    List<String> tableNames = conf.getApplicationName().getTableNames();
    for (String tableName : tableNames) {
      ServiceNowTableInfo tableInfo = getTableMetaData(tableName, conf);
      if (tableInfo == null) {
        continue;
      }
      tableInfos.add(tableInfo);
    }

    return tableInfos;

The list of tables for each application has been hardcoded in the `SourceApplication` enum 

image

The `tableNameField` in the plugin configuration is not being read and thus not required.
More information on the ticket: [PLUGIN-979](https://cdap.atlassian.net/browse/PLUGIN-979)

sgarg-CS:
The `tableNameField` is required in the source code for the Reporting Mode. The default value in the `tablenameField`: _tablename_ is appended to the schema and is being read by the BigQueryMultiTable Sink.

So this issue and PR remains invalid 
Thank you!

PR #8, opened 2021-12-08
Plugin-998 : Develop ServiceNow Batch Multi Source Connector

sgarg-CS:
[PLUGIN-998](https://cdap.atlassian.net/browse/PLUGIN-998)


src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+  }
+
+  /**
+   * Returns the value type chosen.
+   *
+   * @param collector The failure collector to collect the errors
+   * @return An instance of SourceValueType
+   */
+  public SourceValueType getValueType(FailureCollector collector) {
```
vikasrathee-cs, line 213: If this method is not getting used outside the class, make it private

sgarg-CS: No, it's being used by test classes.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -0,0 +1,121 @@
...
+    } catch (Exception e) {
+      LOG.error("Error decoding row from table " + tableName, e);
+      throw new IOException("Error decoding row from table " + tableName, e);
+    }
+    return recordBuilder.build();
+  }
+
+  public void fetchData() {
+    tableName = split.getTableName();
+    tableNameField = multiSourcePluginConf.getTableNameField();
```
vikasrathee-cs, line 86: Add handling for preview run.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -159,7 +162,7 @@ private void fetchSchema(ServiceNowTableAPIClientImpl restApi) {
     schema = Schema.recordOf(tableName, schemaFields);
   }
 
-  private Object convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record) {
+  public Object convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record) {
```
vikasrathee-cs, line 165: use protected instead of public, if you are using it only in child class.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.oltu.oauth2.common.exception.OAuthProblemException;
+import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Set;
+
+import javax.annotation.Nullable;
+
+import static io.cdap.plugin.servicenow.source.util.ServiceNowConstants.HEADER_NAME_TOTAL_COUNT;
```
rmstar, line 40: avoid static import

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
```diff
@@ -0,0 +1,140 @@
...
+            if (apiResponse.getHttpStatus() == HttpStatus.SC_BAD_REQUEST) {
+              collector.addFailure("Bad Request. Table: " + tableName + " is invalid.", "")
+                .withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAMES);
+            }
+          } else if (serviceNowTableAPIClient.parseResponseToResultListOfMap(apiResponse.getResponseBody()).isEmpty()) {
+            collector.addFailure("Table: " + tableName + " is empty.", "")
+              .withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAMES);
+          }
+        } catch (OAuthSystemException | OAuthProblemException e) {
+          LOG.error("Error in fetchTableRecords", e);
```
rmstar, line 133: add errors in failure collector

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -68,7 +72,7 @@ public void initialize(InputSplit split, TaskAttemptContext context) {
   public boolean nextKeyValue() throws IOException {
     try {
       if (results == null) {
-        fetchData();
+         fetchData();
```
rmstar, line 75: revert whitespace change

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
```diff
@@ -16,34 +16,36 @@
...
+import org.apache.oltu.oauth2.common.exception.OAuthProblemException;
+import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.Optional;
 
 import javax.annotation.Nullable;
 
+import static io.cdap.plugin.servicenow.source.util.ServiceNowConstants.HEADER_NAME_TOTAL_COUNT;
```
rmstar, line 42: avoid static import

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
```diff
@@ -395,72 +219,37 @@ private void validateTableQueryMode(FailureCollector collector) {
...
+          if (apiResponse.getHttpStatus() == HttpStatus.SC_BAD_REQUEST) {
+            collector.addFailure("Bad Request. Table: " + tableName + " is invalid.", "")
+              .withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAMES);
+          }
+        } else if (serviceNowTableAPIClient.parseResponseToResultListOfMap(apiResponse.getResponseBody()).isEmpty()) {
+          collector.addFailure("Table: " + tableName + " is empty.", "")
+            .withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAMES);
+        }
+      } catch (OAuthSystemException | OAuthProblemException e) {
+        LOG.error("Error in fetchTableRecords", e);
```
rmstar, line 250: add failure to collector

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -0,0 +1,161 @@
...
+    // Depending on conf value fetch the list of fields for each table and create schema object
+    // return the schema object for each table as ServiceNowTableInfo
+    Set<ServiceNowTableInfo> tableInfos = fetchTablesInfo(conf);
+
+    jobConf.setTableInfos(tableInfos.stream().collect(Collectors.toList()));
+
+    return tableInfos;
+  }
+
+  public static Set<ServiceNowTableInfo> fetchTablesInfo(ServiceNowMultiSourceConfig conf) {
```
rmstar, line 75: why is this method public?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -0,0 +1,161 @@
...
+  public List<InputSplit> getSplits(JobContext jobContext) throws IOException, InterruptedException {
+    ServiceNowJobConfiguration jobConfig = new ServiceNowJobConfiguration(jobContext.getConfiguration());
+
+    List<ServiceNowTableInfo> tableInfos = jobConfig.getTableInfos();
+    List<InputSplit> resultSplits = new ArrayList<>();
+
+    for (ServiceNowTableInfo tableInfo : tableInfos) {
+      String tableName = tableInfo.getTableName();
+      int totalRecords = tableInfo.getRecordCount();
+      if (totalRecords <= ServiceNowConstants.PAGE_SIZE) {
```
rmstar, line 122: you can remove this if condition as the single split case is already handled below

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormatTest.java
```diff
@@ -0,0 +1,73 @@
...
+      "Client Secret",
+      "https://dev115725.service-now.com", "admin", "6qa8xrCJzWTV", "Actual", "2021-12-30", "2021-12-31", "sys_user");
+    ServiceNowColumn column1 = new ServiceNowColumn("sys_created_by", "string");
+    ServiceNowColumn column2 = new ServiceNowColumn("sys_updated_by", "string");
+    List<ServiceNowColumn> columns = new ArrayList<>();
+    columns.add(column1);
+    columns.add(column2);
+    ServiceNowTableDataResponse response = new ServiceNowTableDataResponse();
+    response.setColumns(columns);
+    //when(serviceNowTableAPIClient.fetchTableSchema("sys_user", "2021-12-30","2021-12-31", Boolean.TRUE)).thenReturn
```
rmstar, line 47: remove commented out code

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
```diff
@@ -0,0 +1,140 @@
...
+    return tableNames;
+  }
+
+  /**
+   * Validates {@link ServiceNowMultiSourceConfig} instance.
+   */
+  public void validate(FailureCollector collector) {
+    // Validates the given referenceName to consists of characters allowed to represent a dataset.
+    IdUtils.validateReferenceName(referenceName, collector);
+
```
rmstar, line 87: you can call super.validate() + add multisource specific validations here

sgarg-CS: Refactored the code. Called `super.validate()` and added validation for properties used in Multi Source 7661e08

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormatTest.java
sgarg-CS: Removed. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
sgarg-CS: Removed. Thanks! 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
sgarg-CS: Changed the access modifier to `protected` 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
sgarg-CS: Errors added to the failure collector 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
sgarg-CS: Static Imports have been removed. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: Reverted. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
sgarg-CS: Errors added to the failure collector 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
sgarg-CS: Static Imports have been removed. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -0,0 +1,121 @@
...
+    recordBuilder.set(tableNameField, tableName);
+
+    try {
+      for (Schema.Field field : tableFields) {
+        String fieldName = field.getName();
+        Object fieldValue = convertToValue(fieldName, field.getSchema(), row);
+        recordBuilder.set(fieldName, fieldValue);
+      }
+    } catch (Exception e) {
+      LOG.error("Error decoding row from table " + tableName, e);
```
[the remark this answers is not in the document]
sgarg-CS: Removed the error log. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -0,0 +1,121 @@
...
+
+      if (!iterator.hasNext()) {
+        return false;
+      }
+
+      row = iterator.next();
+
+      pos++;
+    } catch (Exception e) {
+      LOG.error("Error in nextKeyValue", e);
```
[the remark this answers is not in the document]
sgarg-CS: Removed the error log. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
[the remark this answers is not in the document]
sgarg-CS: Updated the year to 2022 for all new class files. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -0,0 +1,161 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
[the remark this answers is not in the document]
sgarg-CS: Updated the year to 2022 for all new class files. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,329 @@
...
+  @Nullable
+  public String getStartDate() {
+    return startDate;
+  }
+
+  public String getEndDate() {
+    return endDate;
+  }
+
+  @Nullable
```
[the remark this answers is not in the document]
sgarg-CS: Yes, we do. Thanks. Removed the @Nullable annotation. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,329 @@
...
+import io.cdap.plugin.servicenow.source.util.SourceValueType;
+import io.cdap.plugin.servicenow.source.util.Util;
+
+import java.time.LocalDate;
+import java.time.temporal.ChronoUnit;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+import static io.cdap.plugin.servicenow.source.util.ServiceNowConstants.TABLE_NAME_FIELD_DEFAULT;
```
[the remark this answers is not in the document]
sgarg-CS: Static Imports have been removed. 7661e08

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -39,7 +39,7 @@
    * @param request the Rest API request
    * @return an instance of RestAPIResponse object.
    */
-  protected RestAPIResponse executeGet(RestAPIRequest request) {
+  public RestAPIResponse executeGet(RestAPIRequest request) {
```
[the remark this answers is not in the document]
sgarg-CS: No, the class calling this method is in a different package than this one. 7661e08

docs/ServiceNowMultiSource-batchsource.md
```diff
@@ -0,0 +1,60 @@
...
+-----------
+
+Reads from one or multiple tables within ServiceNow
+
+Properties
+----------
+
+**Reference Name**: Name used to uniquely identify this source for lineage, annotating metadata, etc.
+
+
```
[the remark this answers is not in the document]
sgarg-CS: Extra newlines have been removed. 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: Changed access modifier from `public` to `protected` 7661e08

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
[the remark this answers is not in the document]
rmstar: Looks like you didn't push this change. The PR was not updated. Same thing for other places where you refer to commit 7661e08.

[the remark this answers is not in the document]
sgarg-CS: I've squashed my commits. Pushed the changes. 6bc4d9e

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /**
+   * Returns the value type chosen.
+   *
+   * @return An instance of SourceValueType
+   */
+  @Nullable
+  public SourceValueType getValueType() {
+    Optional<SourceValueType> sourceValueType = SourceValueType.fromValue(valueType);
```
rmstar, line 231: this can be simplified -
return SourceValueType.fromValue(valueType).orElse(null);

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,327 @@
...
+
+    return sourceValueType.isPresent() ? sourceValueType.get() : null;
+  }
+
+  public void validateValueType(FailureCollector collector) {
+    if (containsMacro(ServiceNowConstants.PROPERTY_VALUE_TYPE)) {
+      return;
+    }
+
```
rmstar, line 240: looks like you aren't really doing any validation here

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
rmstar: Looks like this change didn't get reverted

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -43,16 +43,20 @@
...
+  protected List<Schema.Field> tableFields;
+  protected Schema schema;
+
+  protected String tableName;
+  protected String tableNameField;
+  protected List<Map<String, Object>> results;
+  protected Iterator<Map<String, Object>> iterator;
+  protected Map<String, Object> row;
+
+  public ServiceNowRecordReader() {
```
rmstar, line 57: why do you need this constructor? No need to initialize pluginConf to null

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfigHelper.java
```diff
@@ -43,6 +43,7 @@ public static ConfigBuilder newConfigBuilder() {
     private String applicationName = "";
     private String tableNameField = "tablename";
     private String tableName = "tablename";
+    private String tableNames = "tablesnames";
```
rmstar, line 46: Why do we have 2 fields - tablename and tablenames. Seems confusing

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -0,0 +1,119 @@
...
+    // Fetch the column definition
+    ServiceNowTableDataResponse response = restApi.fetchTableSchema(tableName, null, null,
+      false);
+    if (response == null) {
+      return;
+    }
+
+    // Build schema
+    SchemaBuilder schemaBuilder = new SchemaBuilder();
+    Schema tempSchema = schemaBuilder.constructSchema(tableName, response.getColumns());
```
rmstar, line 111: can this just be a static method?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -185,27 +188,27 @@ private Object convertToValue(String fieldName, Schema fieldSchema, Map<String,
     }
   }
 
-  private String convertToStringValue(Object fieldValue) {
+  public String convertToStringValue(Object fieldValue) {
```
rmstar, line 191: can we make these methods package private?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -0,0 +1,119 @@
...
+
+    LOG.debug("size={}", results.size());
+    if (!results.isEmpty()) {
+      fetchSchema(restApi);
+    }
+
+    iterator = results.iterator();
+  }
+
+  public void fetchSchema(ServiceNowTableAPIClientImpl restApi) {
```
rmstar, line 101: why is this method public

sgarg-CS: Changed access modifier from `public` to `private`.  a7e25f3

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: No for now as they are being used directly in the test class ServiceNowRecordReaderTest.

I will try writing a test case that does not directly call these methods but still covers them in this [PR](https://github.com/data-integrations/servicenow-plugins/pull/9)

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: I'm sorry I couldn't understand. Which method do we want to be static here?

`private void fetchSchema(ServiceNowTableAPIClientImpl restApi) `

**OR**

`public Schema constructSchema(String tableName, List<ServiceNowColumn> columns) `

If it's the latter, then yes. We can make it static.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfigHelper.java
sgarg-CS: `tableName` is to store the table name entered by the user when the mode of  query is set to Table Mode in Service Now Batch Source Plugin
`tableNames` is to store all the table names entered by the user in Service Now Multi-Source Plugin

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: This class `ServiceNowMultiRecordReader` extends `ServiceNowRecordReader` class.

The base class constructor :  
`public ServiceNowRecordReader() {
    pluginConf = null;
  }`

is invoked from the derived class constructor : 

`ServiceNowMultiRecordReader(ServiceNowMultiSourceConfig multiSourcePluginConf) {
    super();
    this.multiSourcePluginConf = multiSourcePluginConf;
  }`
  
  by calling `super()`
  
I had to implement it this way as I was getting a compilation error.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: A macro check is also a form of validation, right? Further there is null check in the method `getValueType(FailureCollector collector)` called from within the method `validateValueType(FailureCollector collector)`

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: Refactored the code.  a7e25f3

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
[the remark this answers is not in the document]
sgarg-CS: I had inadvertently removed it from the `getEndDate()` and not `getTableNameField()`. I've reverted that change. Committed and Pushed.

Please let me know if you're still not able to see it.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,327 @@
+package io.cdap.plugin.servicenow.source;
```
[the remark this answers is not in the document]
sgarg-CS: Copyright added.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
rmstar: hmm then why is ServiceNowMultiRecordReader extending ServiceNowRecordReader? Seems unnecessary if it's not really using methods implemented by ServiceNowRecordReader

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: Changed to package-private and added the annotation: ` @VisibleForTesting` 69415df

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: It is using this method from the base class: `ServiceNowRecordReader.convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record)` 

Also, the return type of the method : `ServiceNowMultiInputFormat.createRecordReader` happens to be an object of class type: `RecordReader<NullWritable, StructuredRecord>` so we have to return a new instance of this class type. 

So I had to either directly extend the `RecordReader` class or extend `ServiceNowRecordReader` class which was further extending `RecordReader` class. i chose the latter as it involved implementing less number of methods.

Please let me know if you feel we can write this piece of code in a better way.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: Yes, Changed it to package-private and added the annotation: ` @VisibleForTesting` 69415df

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Updates the jobConfig with the ServiceNow table information, which will then be read in getSplit() function.
+   *
+   * @param jobConfig the job configuration
+   * @param conf the database conf
+   * @return Collection of ServiceNowTableInfo containing table and schema.
+   */
+  public static Set<ServiceNowTableInfo> setInput(Configuration jobConfig,
+                                                   ServiceNowMultiSourceConfig conf) {
```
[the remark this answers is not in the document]
sgarg-CS: Done 69415df

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,341 @@
...
+    "will be read. Defaults to `tablename`. Note, the Table name field value will be ignored if the Mode " +
+    "is set to `Table`.")
+  protected String tableNameField;
+
+
+  public ServiceNowBaseSourceConfig(String referenceName, String tableNameField, String clientId,
+                                    String clientSecret, String restApiEndpoint,
+                                    String user, String password,
+                                    String valueType, @Nullable String startDate, @Nullable String endDate
+  ) {
```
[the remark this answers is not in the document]
sgarg-CS: Done 69415df

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,341 @@
...
+    return null;
+  }
+
+  /**
+   * Returns the value type chosen.
+   *
+   * @return An instance of SourceValueType
+   */
+  @Nullable
+  public SourceValueType getValueType() {
```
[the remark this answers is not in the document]
sgarg-CS: This method is being referenced from several places within the code.

1. `ServiceNowMultiSourceConfig.validateTableNames(FailureCollector collector)` (Different class, Same package)
2. `ServiceNowSourceConfig.validateTableQueryMode(FailureCollector collector)` (Different class, Same package)
3. `ServiceNowTableAPIClientImpl.fetchTableRecords(String tableName, String startDate, String endDate, int offset,
                                                     int limit)` (Different class, Different package)
4. `ServiceNowTableAPIClientImpl.fetchTableSchemaUsingFirstRecord(String tableName, String startDate,
                                                                       String endDate, boolean fetchRecordCount)` (Different class, Different package)

I can pass in the `collector` argument in the first two usages and make the call to `SourceValueType getValueType(FailureCollector collector)` (the one having formal parameter) instead of `public SourceValueType getValueType()` but for the last two usages, I do not have the `collector` object with me to pass as an argument. I'll have to pass `null` or create a new `FailureCollector` object and then pass it as an actual parameter.

Let me know what you think.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,341 @@
...
+    return SourceValueType.fromValue(valueType).orElse(null);
+  }
+
+  public void validateValueType(FailureCollector collector) {
```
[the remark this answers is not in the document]
sgarg-CS: Changed to `private` 69415df

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,341 @@
...
+    if (containsMacro(ServiceNowConstants.PROPERTY_VALUE_TYPE)) {
+      return;
+    }
+
+    getValueType(collector);
+  }
+
+  public void validateDateRange(FailureCollector collector) {
```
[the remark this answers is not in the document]
sgarg-CS: Changed to `private` 69415df

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -0,0 +1,156 @@
...
+    // Depending on conf value fetch the list of fields for each table and create schema object
+    // return the schema object for each table as ServiceNowTableInfo
+    Set<ServiceNowTableInfo> tableInfos = fetchTablesInfo(conf);
+
+    jobConf.setTableInfos(tableInfos.stream().collect(Collectors.toList()));
+
+    return tableInfos;
+  }
+
+  protected static Set<ServiceNowTableInfo> fetchTablesInfo(ServiceNowMultiSourceConfig conf) {
```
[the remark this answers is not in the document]
sgarg-CS: I've changed it to `package-private` 69415df

This method is being referenced from two places.

1. `ServiceNowMultiInputFormat.setInput(Configuration jobConfig,
                                                  ServiceNowMultiSourceConfig conf)` (From within the class)

2. `ServiceNowMultiSource.configurePipeline(PipelineConfigurer pipelineConfigurer)`  (From a different class within the same package)

sgarg-CS:
@rmstar 

To resolve the default constructor issue,
I've created an abstract class: `ServiceNowBaseRecordReader` with a basic structure having common methods. This class will a base class and will be used by `ServiceNowRecordReader` and `ServiceNowMultiRecordReader` classes.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
rmstar: The latter. I think there's any reason to create an instance of SchemaBuilder - it's just a util class

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfigHelper.java
rmstar: I see.. please add a comment indicating what these fields are used for

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -49,9 +50,9 @@
   private static final String HEADER_NAME_TOTAL_COUNT = "X-Total-Count";
```
rmstar, line 50: looks like this is also defined in ServiceNowConstants. can we dedupe?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
```diff
@@ -395,72 +212,44 @@ private void validateTableQueryMode(FailureCollector collector) {
     if (Util.isNullOrEmpty(tableName)) {
       collector.addFailure("Table name must be specified.", null)
         .withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAME);
+    } else {
+      // Call API to fetch first record from the table
+      ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
```
rmstar, line 217: looks like we have the same config validation logic in ServiceNowMultiSourceCOnfig. Can we dedupe this?

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfigHelper.java
sgarg-CS: Comment added. [c5d6fbb](https://github.com/data-integrations/servicenow-plugins/pull/8/commits/c5d6fbb1b2d49b43e78aadd471064b92574c868e)

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: I've made this method: `Schema constructSchema(String tableName, List<ServiceNowColumn> columns)` static and refactored the method call. Now, it's using the class name to directly call the method instead of creating a new instance of that class. [c5d6fbb](https://github.com/data-integrations/servicenow-plugins/pull/8/commits/c5d6fbb1b2d49b43e78aadd471064b92574c868e)

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
sgarg-CS: Yes, Removed the duplicates. Defined the validation logic in the base class: `ServiceNowBaseSourceConfig` [c5d6fbb](https://github.com/data-integrations/servicenow-plugins/pull/8/commits/c5d6fbb1b2d49b43e78aadd471064b92574c868e)

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Yes, removed the duplicates. [c5d6fbb](https://github.com/data-integrations/servicenow-plugins/pull/8/commits/c5d6fbb1b2d49b43e78aadd471064b92574c868e)

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReaderTest.java
```diff
@@ -0,0 +1,143 @@
...
+  @Test
+  public void testConstructor2() throws IOException {
+
+    serviceNowRecordReader.close();
+    Assert.assertEquals(0, serviceNowRecordReader.pos);
+    Assert.assertEquals("pipeline.user.1", serviceNowSourceConfig.getUser());
+    Assert.assertEquals("tablename", serviceNowSourceConfig.getTableNameField());
+    Assert.assertEquals("tablename", serviceNowSourceConfig.getTableName());
+    Assert.assertEquals("2021-12-30", serviceNowSourceConfig.getStartDate());
+    Assert.assertEquals("https://ven05127.service-now.com", serviceNowSourceConfig.getRestApiEndpoint());
```
rmstar, line 90: do these tests only work against a specific salesforce instance (i.e. your instance with your credentials)?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSource.java
```diff
@@ -0,0 +1,126 @@
...
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+
+    LOG.debug("Validate config during `configurePipeline` stage: {}", conf);
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    FailureCollector collector = stageConfigurer.getFailureCollector();
+
+    conf.validate(collector);
+    collector.getOrThrowException();
```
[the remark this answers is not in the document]
sgarg-CS: I had written this validation logic for tables having no data separately in `ServiceNowBaseSourceConfig.validateTable(String tableName, FailureCollector collector) `

` else if (serviceNowTableAPIClient.parseResponseToResultListOfMap(apiResponse.getResponseBody()).isEmpty()) {
        collector.addFailure("Table: " + tableName + " is empty.", "");
      }`

Have deduped it from `configurePipeline(PipelineConfigurer pipelineConfigurer)` for both `ServiceNowSource` as well as `ServiceNowMultiSource`. 

910499a

Thanks!

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReaderTest.java
sgarg-CS: They'll work against all Service Now instances. The following Assertions:

- `Assert.assertEquals("pipeline.user.1", serviceNowSourceConfig.getUser());`
- `Assert.assertEquals("https://ven05127.service-now.com", serviceNowSourceConfig.getRestApiEndpoint());` 
are dependent on the the user provided VM arguments while running the test case. If the system properties such has _CLIENT_ID, CLIENT_SECRET, REST_API_ENDPOINT, USER, PASSWORD_  are not provided in the VM arguments during test case run, then the test case gets ignored.

If you'd like, I can have these 2 assertions removed.

sgarg-CS: I've removed all test cases and assertions that were dependent upon user credentials.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfigTest.java
```diff
@@ -0,0 +1,180 @@
...
+        .setTableNameField("tablename")
+        .buildMultiSource();
+    MockFailureCollector mockFailureCollector = new MockFailureCollector("Stage Name");
+    serviceNowMultiSourceConfig.validateTableNames(mockFailureCollector);
+    List<ValidationFailure> validationFailures = mockFailureCollector.getValidationFailures();
+    Assert.assertEquals(1, validationFailures.size());
+    ValidationFailure getResult = validationFailures.get(0);
+    List<ValidationFailure.Cause> causes = getResult.getCauses();
+    Assert.assertEquals(0, causes.size());
+    Assert.assertEquals("Table: clm_contract_history is empty.", getResult.getMessage());
```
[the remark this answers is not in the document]
sgarg-CS: > Is this table always going to be empty on every ServiceNow instance?

Yes, by default, whenever a new instance is requested, this table is empty.


> General comment for all tests in this PR: instead of asking the user to provide ServiceNow credentials and connecting to a real instance, prefer to mock out API calls in the test.```

I've removed all instances wherever the tests required the user to specify ServiceNow credentials. [d5a1fe](https://github.com/data-integrations/servicenow-plugins/pull/8/commits/d5a1fe372863518cbd195206cb8ddcf0aac2e4a4)

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfigTest.java
```diff
@@ -0,0 +1,123 @@
...
+    Assert.assertEquals("2021-12-30", serviceNowMultiSourceConfig.getStartDate());
+    Assert.assertEquals("2021-12-31", serviceNowMultiSourceConfig.getEndDate());
+    Assert.assertEquals("tablename", serviceNowMultiSourceConfig.getTableNameField());
+  }
+
+  @Test
+  public void testValidate() {
+    MockFailureCollector mockFailureCollector = new MockFailureCollector("Stage Name");
+    serviceNowMultiSourceConfig.validate(mockFailureCollector);
+    Assert.assertEquals(3, mockFailureCollector.getValidationFailures().size());
```
rmstar, line 69: why are there 3 failures? Generally test 1 thing at a time. It should be clear what you're testing from reading the test code. Also, prefer the initialize serviceNowMultiSourceConfig in the same method for better readability

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormatTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Test
+  public void testFetchTablesInfo() {
+    ServiceNowMultiSourceConfig config =
+      new ServiceNowMultiSourceConfig("Reference Name", "tableNameTable", "client_id",
+                                      "client_secret", "http://example.com", "user",
+                                      "password", "Actual", "2021-12-30",
+                                      "2021-12-31", "sys_user");
+    ServiceNowColumn column1 = new ServiceNowColumn("sys_created_by", "string");
+    ServiceNowColumn column2 = new ServiceNowColumn("sys_updated_by", "string");
```
rmstar, line 37: why are no table infos fetched? If the plugin connects to ServiceNow to fetch table info, you probably want to mock that out and return some fake values for testing purposes.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReaderTest.java
```diff
@@ -0,0 +1,123 @@
...
+  @Test
+  public void testConvertToValue() {
+    Schema fieldSchema = Schema.of(Schema.LogicalType.TIMESTAMP_MILLIS);
+    thrown.expect(IllegalStateException.class);
+    serviceNowRecordReader.convertToValue("Field Name", fieldSchema, new HashMap<>(1));
+  }
+
+  @Test
+  public void testConvertToValue2() {
+    // TODO: This test is incomplete.
```
rmstar, line 92: address TODO?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: The whitespace change had been reverted.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: The preview run is taking 9:58 minutes for displaying 100 records for tables with approx. 50K records. 
A similar issue was observed in the salesforce-marketing plugin where the following workaround had been proposed to fix the issue. 
https://github.com/data-integrations/salesforce-marketing/pull/16#discussion_r789174283
They are yet to be approved and merged into `develop`.
If we have a general consensus on this fix, then I can go ahead and implement the same for service-now plugins.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfigTest.java
sgarg-CS: The test case is being re-written to have only 1 assertion. `serviceNowMultiSourceConfig` will be initialized in the test method itself. The refactored test case will be included in the next PR for service-now JUnit.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfigTest.java
```diff
@@ -0,0 +1,123 @@
...
+        .buildMultiSource();
+    serviceNowMultiSourceConfig.validate(mockFailureCollector);
+    Assert.assertEquals(4, mockFailureCollector.getValidationFailures().size());
+  }
+
+  @Test
+  public void testValidateTableNames() {
+    MockFailureCollector mockFailureCollector = new MockFailureCollector("Stage Name");
+    serviceNowMultiSourceConfig.validateTableNames(mockFailureCollector);
+    Assert.assertEquals(1, mockFailureCollector.getValidationFailures().size());
```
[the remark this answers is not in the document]
sgarg-CS: There was no connection being made to the Service Now instance and hence a generic failure of **Unable to connect** was being returned. This test case is being refactored and will be a part of the next PR for JUnit.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormatTest.java
sgarg-CS: Rework is in progress to replace the actual rest API call to fetch table info with the mocked one. This will be refactored and part of the next PR for service-now JUnits.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReaderTest.java
sgarg-CS: This has been removed. Will be taken care of, in the next PR for service-now JUnits.

sgarg-CS:
hi @rmstar 

The rework for refactoring the JUnit's and replacing the actual Rest API calls with the mocked ones is currently in progress. [PR](https://github.com/data-integrations/servicenow-plugins/pull/9/) for the JUnit's will be raised next week preferably by Tuesday.

The PR for performance fix (also likely to be raised next week) has a dependency on the current [PR](https://github.com/data-integrations/servicenow-plugins/pull/8). Therefore, I request you to review the current [PR](https://github.com/data-integrations/servicenow-plugins/pull/8) excluding the JUnit's, so that it gets merged to `develop` branch soon.
Please let us know if any further rework is required.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,371 @@
...
+  private String endDate;
+
+  @Name(ServiceNowConstants.PROPERTY_TABLE_NAME_FIELD)
+  @Macro
+  @Nullable
+  @Description("The name of the field that holds the table name. Must not be the name of any table column that " +
+    "will be read. Defaults to `tablename`. Note, the Table name field value will be ignored if the Mode " +
+    "is set to `Table`.")
+  protected String tableNameField;
+
```
rmstar, line 102: nit: remove extra line

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfigTest.java
```diff
@@ -176,136 +160,136 @@ public void testValidateClientIdNull() {
...
       config.validate(collector);
       collector.getOrThrowException();
     } catch (ValidationException e) {
-      Assert.assertEquals(PROPERTY_CLIENT_ID, e.getFailures().get(0).getCauses().get(0)
+      Assert.assertEquals(ServiceNowConstants.PROPERTY_CLIENT_ID, e.getFailures().get(0).getCauses().get(0)
         .getAttribute(CauseAttributes.STAGE_CONFIG));
     }
 
-    Assert.assertEquals(1, collector.getValidationFailures().size());
+    Assert.assertEquals(2, collector.getValidationFailures().size());
```
rmstar, line 177: Why has the number of validation failures changed in existing tests? Similar comment for other places where collector.getValidationFailures().size() has increased

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
```diff
@@ -0,0 +1,128 @@
...
+  String convertToStringValue(Object fieldValue) {
+    return String.valueOf(fieldValue);
+  }
+
+  @VisibleForTesting
+  Double convertToDoubleValue(Object fieldValue) {
+    if (fieldValue instanceof String && Strings.isNullOrEmpty(String.valueOf(fieldValue))) {
+      return null;
+    }
+
```
rmstar, line 107: Should we handle the case where fieldValue is not instanceof String here and in other convertTo* methods?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
```diff
@@ -16,34 +16,26 @@
...
 
 /**
  * Configuration for the {@link ServiceNowSource}.
  */
-public class ServiceNowSourceConfig extends PluginConfig {
-  @Name(Constants.Reference.REFERENCE_NAME)
-  @Description("This will be used to uniquely identify this source for lineage, annotating metadata, etc.")
-  private String referenceName;
+public class ServiceNowSourceConfig extends ServiceNowBaseSourceConfig {
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceNowSourceConfig.class);
```
rmstar: is this unused?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Set;
+
+import javax.annotation.Nullable;
+
+/**
+ * Configuration for the {@link ServiceNowMultiSource}.
+ */
+public class ServiceNowMultiSourceConfig extends ServiceNowBaseSourceConfig {
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceNowMultiSourceConfig.class);
```
rmstar: unused?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSource.java
```diff
@@ -0,0 +1,120 @@
...
+    // Since we have validated all the properties, throw an exception if there are any errors in the collector.
+    // This is to avoid adding same validation errors again in getSchema method call
+    if (conf.shouldGetSchema()) {
+      Set<ServiceNowTableInfo> tableInfos = ServiceNowMultiInputFormat.fetchTablesInfo(conf);
+      stageConfigurer.setOutputSchema(tableInfos.stream().findFirst().get().getSchema());
```
rmstar: why is output schema set to schema of the first table? I think we can just set it to null. Ex - https://github.com/data-integrations/multi-table-plugins/blob/develop/src/main/java/io/cdap/plugin/MultiTableDBSource.java#L84

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSource.java
```diff
@@ -0,0 +1,120 @@
...
+    super.configurePipeline(pipelineConfigurer);
+
+    LOG.debug("Validate config during `configurePipeline` stage: {}", conf);
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    FailureCollector collector = stageConfigurer.getFailureCollector();
+
+    conf.validate(collector);
+    collector.getOrThrowException();
+    // Since we have validated all the properties, throw an exception if there are any errors in the collector.
+    // This is to avoid adding same validation errors again in getSchema method call
```
rmstar: I think this comment applies to the previous line, so you may want to move it up

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -0,0 +1,371 @@
...
+        .withConfigProperty(ServiceNowConstants.PROPERTY_CLIENT_SECRET)
+        .withConfigProperty(ServiceNowConstants.PROPERTY_API_ENDPOINT)
+        .withConfigProperty(ServiceNowConstants.PROPERTY_USER)
+        .withConfigProperty(ServiceNowConstants.PROPERTY_PASSWORD);
+    }
+  }
+
+}
+
+
```
rmstar: nit: remove blank lines

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Name(ServiceNowConstants.PROPERTY_TABLE_NAMES)
+  @Macro
+  @Description("The names of the ServiceNow table from which data to be fetched.")
```
rmstar: table->tables

rmstar (commented):
Can you add a sample pipeline under examples/ dir?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: Removed. 6c4965a

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfigTest.java
sgarg-CS: The first failure in the collector is **Client ID must be specified.** which is due to `ClientId` being set as `null` while the second failure in the collector is **Unable to connect to ServiceNow Instance.**
A new validation was added to check for invalid or empty tables in `ServiceNowBaseSourceConfig.validateTable(String tableName, FailureCollector collector)`.
As there is no mock call to ServiceNow, the test case tries to make an actual call to the ServiceNow instance and gets a failure.

This JUnit will be refactored in the Next PR.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
sgarg-CS: Before checking for `null` or empty value and while parsing the `String` value to Double value, the `fieldValue` is already being converted to `String` value by this piece of code: `Strings.isNullOrEmpty(String.valueOf(fieldValue))` and `Double.parseDouble(String.valueOf(fieldValue));` respectively, so handling the case where we receive a fieldValue which does not happen to be an instance of String shouldn't be required in my opinion.

That being said, if there's any specific use case that you believe, which is not covered here, we can take care of it.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSourceConfig.java
sgarg-CS: Yep. removed. 6c4965a

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
sgarg-CS: Yep, removed. 6c4965a

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSource.java
sgarg-CS: CDAP currently doesn't support displaying schema for all tables, if there is more than 1 table. So, it was mutually agreed that the schema of the first table entered by the user gets displayed as the output schema.
As per the reference shared by you for `MultiTableDBSource`, I've set the output schema to `null`. 6c4965a

I have one query here.
Since the Service Now Batch Source Plugin Reporting Mode's  behaviour is similar to Service Now Batch Multi-Source Plugin and has output schema set to the schema of the first table, should we set the output schema to display `null` there as well?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSource.java
sgarg-CS: Moved 1 line up. 6c4965a

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: Removed. 6c4965a

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSourceConfig.java
sgarg-CS: Done. 6c4965a

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: > Can you add a sample pipeline under examples/ dir?

Added. feecb3f

rmstar (approved):
please squash commits

sgarg-CS:
> please squash commits

Commits have been squashed. 60f3119646685dbdea0cd1072bc6abcd56a345d5
Thanks, Prashant.

PR #16, opened 2022-03-14
PLUGIN-1089: Improve ServiceNow Plugins Performance

sgarg-CS:
[PLUGIN-1089](https://cdap.atlassian.net/browse/PLUGIN-1089) 

Add retry to ServiceNow RestAPI Calls when you get the maximum execution time exceeded error or `OauthSystemException`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.github.rholder</groupId>
+      <artifactId>guava-retrying</artifactId>
+      <version>2.0.0</version>
+    </dependency>
+    <dependency>
       <groupId>org.powermock</groupId>
       <artifactId>powermock-module-junit4</artifactId>
       <version>${powermock.version}</version>
--- a/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
+++ b/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
@@ -16,6 +16,9 @@
 
 package io.cdap.plugin.servicenow.restapi;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import io.cdap.plugin.servicenow.source.util.ServiceNowConstants;
 import org.apache.http.Header;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
@@ -48,11 +51,13 @@ public class RestAPIResponse {
   private int httpStatus;
   private Map<String, String> headers;
   private String responseBody;
+  private boolean isRetryable;
 
   public RestAPIResponse(int httpStatus, Map<String, String> headers, String responseBody) {
     this.httpStatus = httpStatus;
     this.headers = headers;
     this.responseBody = responseBody;
+    this.checkRetryable();
   }
 
   public static RestAPIResponse defaultErrorResponse(String message) {
@@ -98,7 +103,26 @@ public class RestAPIResponse {
   }
 
   public boolean isSuccess() {
-    return successCodes.contains(getHttpStatus());
+    boolean isSuccess = false;
+    // ServiceNow Rest API may return 200 OK response with an error message in the response body.
+    // Normally we would expect non-200 response code if there's an error.
+    if (this.isRetryable) {
+      isSuccess = false;
+    } else if (successCodes.contains(getHttpStatus())) {
+      isSuccess = true;
+    }
+    return isSuccess;
+  }
+
+  private void checkRetryable() {
+    Gson gson = new Gson();
+    JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
+    if (jo.get(ServiceNowConstants.STATUS) != null &&
+      jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
+      jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
+        .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+      isRetryable = true;
+    }
   }
 
   public Map<String, String> getHeaders() {
@@ -107,5 +131,9 @@ public class RestAPIResponse {
 
   public String getResponseBody() {
     return responseBody;
+  }
+
+  public boolean isRetryable() {
+    return isRetryable;
   }
 }
--- a/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.servicenow.source;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.plugin.servicenow.source.apiclient.ServiceNowTableAPIClientImpl;
@@ -78,18 +79,17 @@ public class ServiceNowMultiRecordReader extends ServiceNowBaseRecordReader {
     return recordBuilder.build();
   }
 
-  public void fetchData() {
+  @VisibleForTesting
+  void fetchData() {
     tableName = split.getTableName();
     tableNameField = multiSourcePluginConf.getTableNameField();
 
     ServiceNowTableAPIClientImpl restApi = new ServiceNowTableAPIClientImpl(multiSourcePluginConf);
 
     // Get the table data
-    results =
-      restApi.fetchTableRecords(tableName, multiSourcePluginConf.getStartDate(), multiSourcePluginConf.getEndDate(),
-                                split.getOffset(), ServiceNowConstants.PAGE_SIZE);
-
-    LOG.debug("size={}", results.size());
+    results = restApi.fetchTableRecordsRetryableMode(tableName, multiSourcePluginConf.getStartDate(),
+      multiSourcePluginConf.getEndDate(), split.getOffset(), ServiceNowConstants.PAGE_SIZE);
+    LOG.debug("Results size={}", results.size());
     if (!results.isEmpty()) {
       fetchSchema(restApi);
     }
--- a/src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
@@ -16,6 +16,11 @@
 
 package io.cdap.plugin.servicenow.source.apiclient;
 
+import com.github.rholder.retry.RetryException;
+import com.github.rholder.retry.Retryer;
+import com.github.rholder.retry.RetryerBuilder;
+import com.github.rholder.retry.StopStrategies;
+import com.github.rholder.retry.WaitStrategies;
 import com.google.common.base.Strings;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
@@ -37,6 +42,9 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Implementation class for ServiceNow Table API.
@@ -48,9 +56,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   private static final String FIELD_CREATED_ON = "sys_created_on";
   private static final String FIELD_UPDATED_ON = "sys_updated_on";
   private static final String OAUTH_URL_TEMPLATE = "%s/oauth_token.do";
-
   private ServiceNowBaseSourceConfig conf;
-
+  
   public ServiceNowTableAPIClientImpl(ServiceNowBaseSourceConfig conf) {
     this.conf = conf;
   }
@@ -91,12 +98,16 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
       requestBuilder.setAuthHeader(accessToken);
       apiResponse = executeGet(requestBuilder.build());
       if (!apiResponse.isSuccess()) {
-        LOG.error("Error - {}", getErrorMessage(apiResponse.getResponseBody()));
+        if (apiResponse.isRetryable()) {
+          throw new RetryableException();
+        }
         return Collections.emptyList();
       }
 
       return parseResponseToResultListOfMap(apiResponse.getResponseBody());
-    } catch (OAuthSystemException | OAuthProblemException e) {
+    } catch (OAuthSystemException e) {
+      throw new RetryableException();
+    } catch (OAuthProblemException e) {
       LOG.error("Error in fetchTableRecords", e);
       return Collections.emptyList();
     }
@@ -198,12 +209,12 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
 
   public List<Map<String, Object>> parseResponseToResultListOfMap(String responseBody) {
     Gson gson = new Gson();
+
     JsonObject jo = gson.fromJson(responseBody, JsonObject.class);
     JsonArray ja = jo.getAsJsonArray("result");
 
     Type type = new TypeToken<List<Map<String, Object>>>() {
     }.getType();
-
     return gson.fromJson(ja, type);
   }
 
@@ -215,5 +226,40 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
     } catch (Exception e) {
       return e.getMessage();
     }
+  }
+
+  /**
+   * Attempt four times with an exponential delay of 120 seconds to fetch the list of records from ServiceNow table when
+   * RetryableException is thrown          .
+   *
+   * @param tableName The ServiceNow table name
+   * @param startDate The start date
+   * @param endDate The end date
+   * @param offset The number of records to skip
+   * @param limit The number of records to be fetched
+   * @return The list of Map; each Map representing a table row
+   */
+  public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName, String startDate, String endDate,
+                                                                  int offset, int limit) {
+    final List<Map<String, Object>> results = new ArrayList<>();
+    Callable<Boolean> fetchRecords = () -> {
+      results.addAll(fetchTableRecords(tableName, startDate, endDate, offset, limit));
+      return true;
+    };
+
+    Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
+      .retryIfExceptionOfType(RetryableException.class)
+      .withWaitStrategy(WaitStrategies.exponentialWait(ServiceNowConstants.WAIT_TIME, TimeUnit.MILLISECONDS))
+      .withStopStrategy(StopStrategies.stopAfterAttempt(ServiceNowConstants.MAX_NUMBER_OF_RETRY_ATTEMPTS))
+      .build();
+
+    try {
+      retryer.call(fetchRecords);
+    } catch (RetryException | ExecutionException e) {
+      LOG.error("Data Recovery failed for batch {} to {}.", offset,
+               (offset + limit));
+    }
+
+    return results;
   }
 }
--- a/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowInputFormat.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowInputFormat.java
@@ -112,7 +112,7 @@ public class ServiceNowInputFormat extends InputFormat<NullWritable, StructuredR
   }
 
   @Override
-  public List<InputSplit> getSplits(JobContext jobContext) throws IOException, InterruptedException {
+  public List<InputSplit> getSplits(JobContext jobContext) {
     ServiceNowJobConfiguration jobConfig = new ServiceNowJobConfiguration(jobContext.getConfiguration());
 
     List<ServiceNowTableInfo> tableInfos = jobConfig.getTableInfos();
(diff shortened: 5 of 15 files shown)


sgarg-CS:
hi @rmstar 

Can you please review this PR?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -48,6 +50,7 @@
   private int httpStatus;
   private Map<String, String> headers;
   private String responseBody;
+  private boolean isRetriable;
```
vikasrathee-cs: rename to retryable

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -98,7 +101,24 @@ public int getHttpStatus() {
...
+    } else if (successCodes.contains(getHttpStatus())) {
+      isSuccess = true;
+    }
+    return isSuccess;
+  }
+
+  private boolean checkRetriable() {
+    Gson gson = new Gson();
+    JsonObject jo = gson.fromJson(this.getResponseBody(), JsonObject.class);
+    if (jo.get("status") != null && jo.get("status").getAsString().equals("failure")) {
```
vikasrathee-cs: define keywords like status failure or error message as String literals and use them.
Merge these two for loops into one.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -78,23 +78,25 @@ public StructuredRecord getCurrentValue() throws IOException {
...
     // Get the table data
-    results =
-      restApi.fetchTableRecords(tableName, multiSourcePluginConf.getStartDate(), multiSourcePluginConf.getEndDate(),
-                                split.getOffset(), ServiceNowConstants.PAGE_SIZE);
-
-    LOG.debug("size={}", results.size());
+    results = restApi.fetchTableRecordsRetriableMode(tableName, multiSourcePluginConf.getStartDate(),
+      multiSourcePluginConf.getEndDate(), split.getOffset(), ServiceNowConstants.PAGE_SIZE, false);
     if (!results.isEmpty()) {
+      LOG.debug("size={}", results.size());
```
vikasrathee-cs: write a more clear message for user to understand.

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -48,7 +49,14 @@
   private static final String FIELD_CREATED_ON = "sys_created_on";
   private static final String FIELD_UPDATED_ON = "sys_updated_on";
   private static final String OAUTH_URL_TEMPLATE = "%s/oauth_token.do";
-
+  private static final Map<Integer, Integer> retryDelay = new HashMap<Integer, Integer>() {
```
vikasrathee-cs: make it immutable.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Renamed to `retryable` here and at other places. a752f6b05f68ae64ce9ed63c94fc7b1cc1865b12

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Defined String Literals and merged two **if** blocks a752f6b05f68ae64ce9ed63c94fc7b1cc1865b12

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: Re-written the debug logs that print the size of the result. a752f6b05f68ae64ce9ed63c94fc7b1cc1865b12

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Replaced the HashMap with ImmutableMap. a752f6b05f68ae64ce9ed63c94fc7b1cc1865b12

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -48,7 +49,11 @@
   private static final String FIELD_CREATED_ON = "sys_created_on";
   private static final String FIELD_UPDATED_ON = "sys_updated_on";
   private static final String OAUTH_URL_TEMPLATE = "%s/oauth_token.do";
-
+  private static final ImmutableMap<Integer, Integer> retryDelay = ImmutableMap.of(
```
rmstar: Use a library instead of implementing retry/backoff logic yourself.
Example - https://github.com/data-integrations/google-cloud/pull/918/files#diff-6a7f8b65e9e06b9f3d93abd2603ec319c4e9c5b6083fcadac773b32f55e1da8cR318

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -89,14 +94,23 @@ public String getAccessToken() throws OAuthSystemException, OAuthProblemExceptio
     try {
       String accessToken = getAccessToken();
       requestBuilder.setAuthHeader(accessToken);
+      LOG.debug("Request URL: {} ", requestBuilder.build().getUrl());
+      long start = System.currentTimeMillis();
       apiResponse = executeGet(requestBuilder.build());
+      long end = System.currentTimeMillis();
+      LOG.debug("restAPI execution time for {} to {} records took {}s ", offset,
```
rmstar: we can remove logging added for debugging purposes

rmstar: we can remove logging added for debugging purposes

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -98,7 +102,25 @@ public int getHttpStatus() {
   }
 
   public boolean isSuccess() {
-    return successCodes.contains(getHttpStatus());
+    boolean isSuccess = false;
+    if (checkRetryable()) {
+      isSuccess = false;
```
rmstar, line 110: why do we need need this if (checkRetryable()) check?  Can't we just return true if successCodes.contains(getHttpStatus()) - i.e. what this method was doing previously

rmstar, line 110: why do we need need this if (checkRetryable()) check?  Can't we just return true if successCodes.contains(getHttpStatus()) - i.e. what this method was doing previously

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -108,4 +130,8 @@ public boolean isSuccess() {
   public String getResponseBody() {
     return responseBody;
   }
+  
+  public Boolean getRetryable() {
```
rmstar: why do we need 2 method - checkRetryable and getRetryable.

rmstar: why do we need 2 method - checkRetryable and getRetryable.

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Removed the logger ecfef61ba226b88cb299cac832d75aa7624808ae

sgarg-CS: Removed the logger ecfef61ba226b88cb299cac832d75aa7624808ae

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Previously this method: `public boolean isSuccess()` was returning true if the HTTP response code was 200 OK.

But there was no logic in place for API calls taking more execution time than the specified limit (60 sec). Here, along with the fetched data, we get an error message saying **Error - com.glide.sys.TransactionCancelledException: Transaction cancelled: maximum execution time exceeded** in the response JSON for all REST API calls that take more than 60 sec and eventually lose that batch. [Data Loss Issue]

To overcome this, we added the retry functionality. Whenever this error is encountered, the connector will retry the API. The `checkRetryable()` method contains the logic as to when the retry functionality should be invoked.

sgarg-CS: Previously this method: `public boolean isSuccess()` was returning true if the HTTP response code was 200 OK.

This included the scenario where the ServiceNow Table Rest API returned a 200 OK response for API calls taking more execution time than the specified limit (60 sec). Here, we get an error message saying Error - **com.glide.sys.TransactionCancelledException: Transaction cancelled: maximum execution time exceeded** in the response JSON for all REST API calls that take more than 60 sec and eventually lose that batch. [Data Loss Issue]

To overcome this, we added the retry functionality. Whenever this error is encountered, the connector will retry the API. The `checkRetryable()` method contains the logic as to when the retry functionality should be invoked.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: The first method : `checkRetryable` is to check whether the retry is required or not. It sets the `isRetryable` variable to true when the error message : _maximum execution time exceeded_ is encountered.
The second method `getRetryable` returns the value of the class level variable: `isRetryable`. It's a getter method.

sgarg-CS: The first method : `checkRetryable` is to check whether the retry is required or not. It sets the `isRetryable` variable to true when the error message : _maximum execution time exceeded_ is encountered.
The second method `getRetryable` returns the value of the class level variable: `isRetryable`. It's a getter method.

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Couldn't find the reference to the library related to retry on the link shared by you.

Can we have a short call for this?

rmstar: Try using https://github.com/rholder/guava-retrying.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
rmstar: I see, please add a comment indicating that ServiceNow API is returning OK response with an error msg in response body. Normally we would expect non-200 response code if there's an error

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -89,14 +94,20 @@ public String getAccessToken() throws OAuthSystemException, OAuthProblemExceptio
     try {
       String accessToken = getAccessToken();
       requestBuilder.setAuthHeader(accessToken);
+      long start = System.currentTimeMillis();
       apiResponse = executeGet(requestBuilder.build());
+      long end = System.currentTimeMillis();
```
rmstar: start, end are now used. can be removedd

sgarg-CS: `start`, `end` removed. 286b28256112b49eaab4a3a036a86973a3fa3bda

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Added the comment. 76c21ca228efde19affd50f83ad5c0a183bba927

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Thank you for sharing the library. The dev team is done with the implementation using the guava-retry library and is currently unit testing the code. This will be followed by functional testing by the QA team which will likely take a day or two more to complete. 

As this PR, containing a fix for the data loss issue happens to be a time-sensitive deliverable, part of the ServiceNow Release, can we raise a separate PR for the guava-retry implementation and let this one get through?

sgarg-CS: Retry implemented using Guava-retrying library [150737e](https://github.com/data-integrations/servicenow-plugins/pull/16/commits/150737e8afe7681df7a7356fea149043d9c98075)

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -216,4 +228,28 @@ private String getErrorMessage(String responseBody) {
       return e.getMessage();
     }
   }
+
+  public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName, String startDate, String endDate,
```
rmstar, line 242: add javadoc for public method

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -216,4 +228,28 @@ private String getErrorMessage(String responseBody) {
...
+                                                                  int offset, int limit) {
+    final List<Map<String, Object>> results = new ArrayList<>();
+    Callable<Boolean> fetchRecords = () -> {
+      results.addAll(fetchTableRecords(tableName, startDate, endDate, offset, limit));
```
rmstar, line 246: fetchTableRecords can now be private

pom.xml
```diff
@@ -215,7 +215,6 @@
       <version>2.24.0</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
rmstar, line 218: why was this line removed?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -98,7 +102,27 @@ public int getHttpStatus() {
...
+      isSuccess = false;
+    } else if (successCodes.contains(getHttpStatus())) {
+      isSuccess = true;
+    }
+    return isSuccess;
+  }
+
+  private boolean checkRetryable() {
+    Gson gson = new Gson();
+    JsonObject jo = gson.fromJson(this.getResponseBody(), JsonObject.class);
```
rmstar: getRetryable() will return false incorrectly (for retryble error) if isSuccess() was not called previously on the same RestAPIResponse. We shouldn't have such dependency between getters. You can fix this by moving the checkRetryable logic into the constructor.

pom.xml
sgarg-CS: This one got removed while resolving merge conflicts. 
Added back. https://github.com/data-integrations/servicenow-plugins/pull/16/commits/94cb7501d6219ca8756a6122b6daa510c0b7c1b8

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Added. https://github.com/data-integrations/servicenow-plugins/pull/16/commits/94cb7501d6219ca8756a6122b6daa510c0b7c1b8

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: This method is being used by the test classes in a different package.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Moved the `checkRetryable()` call to the constructor and removed the dependency. https://github.com/data-integrations/servicenow-plugins/pull/16/commits/94cb7501d6219ca8756a6122b6daa510c0b7c1b8

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -98,7 +105,27 @@ public int getHttpStatus() {
...
+    // Normally we would expect non-200 response code if there's an error.
+    if (getRetryable()) {
+      throw new RetryableException();
+    } else if (successCodes.contains(getHttpStatus())) {
+      isSuccess = true;
+    }
+    return isSuccess;
+  }
+
+  private boolean checkRetryable() {
```
rmstar: Return value is not used, can return void

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -98,7 +105,27 @@ public int getHttpStatus() {
   }
 
   public boolean isSuccess() {
-    return successCodes.contains(getHttpStatus());
+    boolean isSuccess = false;
+    // ServiceNow Rest API may return 200 OK response with an error message in the response body.
+    // Normally we would expect non-200 response code if there's an error.
+    if (getRetryable()) {
```
rmstar: no need to call the getter, just access isRetryable directly. Similar comment elsewhere (getResponseBody()).

sgarg-CS: Removed the getter : `getRetryable()` 2c4538f24a45de8576bb7a51b2422ad070a696dd

`getResponseBody()` is being used in several places outside the`RestAPIResponse` class

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Changed the return type to void. 2c4538f24a45de8576bb7a51b2422ad070a696dd

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -216,4 +224,39 @@ private String getErrorMessage(String responseBody) {
...
+      return true;
+    };
+
+    Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
+      .retryIfExceptionOfType(RetryableException.class)
+      .withWaitStrategy(WaitStrategies.exponentialWait(120000, TimeUnit.MILLISECONDS))
```
rmstar: Make the wait time and stopAfterAttempt values constants

rmstar (approved):
lgtm. Please squash commits after addressing minor comment

src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Added constants `WAIT_TIME` and `MAX_NUMBER_OF_RETRY_ATTEMPTS`

sgarg-CS:
Commits have been squashed. Thanks!

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -98,7 +104,26 @@ public int getHttpStatus() {
   }
 
   public boolean isSuccess() {
-    return successCodes.contains(getHttpStatus());
+    boolean isSuccess = false;
+    // ServiceNow Rest API may return 200 OK response with an error message in the response body.
+    // Normally we would expect non-200 response code if there's an error.
+    if (this.isRetryable) {
+      throw new RetryableException();
```
rmstar: Sorry, I just noticed this. We shouldn't be throwing an exception in this method - just return true/false.
return isRetryable || successCodes.contains(getHttpStatus());

In the caller, if isSuccess() returns false, throw RetryableException if the error is retryable (add a getter for isRetryable)

rmstar (changes requested):
Sorry for the delay on this. You can squash your commits again after addressing the comment.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Removed the exception from `isSuccess()` method

Added a getter `isRetryable()` in `RestAPIResponse` class. Using this getter to check the value of `isRetryable`  Throwing `RetryableException` if the value is `true`

This is how it looks now:
```
public boolean isSuccess() {
   boolean isSuccess = false;
   // ServiceNow Rest API may return 200 OK response with an error message in the response body.
   // Normally we would expect non-200 response code if there's an error.
   if (this.isRetryable) {
          isSuccess = false;
   } else if (successCodes.contains(getHttpStatus())) {
          isSuccess = true;
   }
   return isSuccess;
}
```

If we use this : `return isRetryable || successCodes.contains(getHttpStatus());`, then when `isRetryable`is `true`, the `isSuccess()` would be true and the below `if` won't get executed and thus retry logic won't be invoked.

```
if (!apiResponse.isSuccess()) {
        if (apiResponse.isRetryable()) {
          throw new RetryableException();
        }
        return Collections.emptyList();
      }
```

sgarg-CS:
Commits squashed. Thanks!

rmstar (approved):

PR #19, opened 2022-03-22
PLUGIN-1142: Display output schema only for Table Mode in ServiceNow Batch Source Plugin

sgarg-CS:
[PLUGIN-1142](https://cdap.atlassian.net/browse/PLUGIN-1142)

~**ServiceNow Multi-Source:** Display output schema only when data is requested for one table. No schema would be displayed if data is requested for more than one table~

**ServiceNow Batch Source:** Display output schema only for Table mode. No schema would be displayed for Reporting Mode

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSource.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowSource.java
@@ -75,9 +75,11 @@ public class ServiceNowSource extends BatchSource<NullWritable, StructuredRecord
     // Since we have validated all the properties, throw an exception if there are any errors in the collector.
     // This is to avoid adding same validation errors again in getSchema method call
     collector.getOrThrowException();
-    if (conf.shouldGetSchema()) {
+    if (conf.shouldGetSchema() && conf.getQueryMode() == SourceQueryMode.TABLE) {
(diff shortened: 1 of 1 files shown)


sgarg-CS:
hi @rmstar 

Please review the changes in this PR.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiSource.java
```diff
@@ -75,7 +75,14 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     // Since we have validated all the properties, throw an exception if there are any errors in the collector.
     // This is to avoid adding same validation errors again in getSchema method calls
     collector.getOrThrowException();
-    stageConfigurer.setOutputSchema(null);
```
rmstar: We can keep the same behavior as other multi source plugins . ex - https://github.com/data-integrations/multi-table-plugins/blob/develop/src/main/java/io/cdap/plugin/MultiTableDBSource.java#L84
Revert the change in this file.

sgarg-CS: Reverted 5b4f0c9cb9d3fb2d99126c231f4f9283f566e69a

rmstar (approved):
please squash commits

sgarg-CS:
Commits squashed. Thanks!

PR #34, opened 2022-08-09
CEC-977 : ServiceNow Sink Run Time Implementation

sgarg-CS:


pom.xml
```diff
@@ -60,7 +60,8 @@
     <main.basedir>${project.basedir}</main.basedir>
     <jee.version>7</jee.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <cdap.version>6.1.1</cdap.version>
+    <cdap.version>6.6.0</cdap.version>
+    <cdap.data.pipeline.version>6.4.1</cdap.data.pipeline.version>
```
rmstar: why is this set to a different version than cdap.version?

pom.xml
```diff
@@ -60,7 +60,8 @@
     <main.basedir>${project.basedir}</main.basedir>
     <jee.version>7</jee.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <cdap.version>6.1.1</cdap.version>
+    <cdap.version>6.6.0</cdap.version>
```
rmstar: Don't bump up cdap.version unless we actually need to. i.e. we're adding a dependency on a CDAP API that's only present in a more recent version

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -70,21 +78,42 @@ public String getAccessToken() throws OAuthSystemException, OAuthProblemExceptio
       conf.getClientSecret(), conf.getUser(), conf.getPassword());
   }
 
+  /**
+   * Retries to get the access token and returns the same when OAuthSystemException is thrown
+   */
+  public String getAccessTokenRetryableMode() throws ExecutionException, RetryException {
+
+    Callable<Boolean> fetchToken = () -> {
```
rmstar: Make this Callable<String> and return accessToken instead of returning an unused boolean

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -59,8 +65,10 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   private static final String FIELD_CREATED_ON = "sys_created_on";
   private static final String FIELD_UPDATED_ON = "sys_updated_on";
   private static final String OAUTH_URL_TEMPLATE = "%s/oauth_token.do";
-  private ServiceNowBaseConfig conf;
-  
+  private static final Gson gson = new Gson();
+  private final ServiceNowBaseConfig conf;
+  private String accessToken;
```
rmstar: not needed, return access token from getAccessTokenRetryableMode()

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -121,11 +150,11 @@ public List<Map<String, Object>> fetchTableRecords(String tableName, SourceValue
    * Create a new record in the ServiceNow Table
    *
    * @param tableName ServiceNow Table name
-   * @param entity Details of the Record to be created
+   * @param entity    Details of the Record to be created
    */
-  public void createRecord(String tableName, HttpEntity entity) {
+  public void createRecord(String tableName, HttpEntity entity) throws IOException {
```
rmstar, line 150: why do we throw IOException now?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -295,9 +323,54 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
+        LOG.error("Error - {}", getErrorMessage(apiResponse.getResponseBody()));
+        collector.addFailure("Unable to fetch schema for table " + tableName, null).
+          withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAME);
+      }
+      List<SchemaResponse> result = parseSchemaResponse(apiResponse.getResponseBody());
+      List<ServiceNowColumn> columns = new ArrayList<>();
+
+      if (result != null && !result.isEmpty()) {
+        for (SchemaResponse field : result) {
+          columns.add(new ServiceNowColumn(field.getName(), "string"));
```
rmstar: are all column types string?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,267 @@
...
+          retryUnservicedRequests(restRequestList, unservicedRequestsArray, accessToken);
+        }
+      }
+    } catch (IOException e) {
+      LOG.error("Error while connecting to ServiceNow", e.getMessage());
+      throw new RetryableException();
+    } catch (Exception e) {
+      throw e;
+    }
+    return apiResponse.getHttpStatus() == HttpStatus.SC_OK;
```
rmstar: just throw an exception if you get an unexpected return code - you're already throwing exception in other failure cases. This method can return void.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,267 @@
...
+  private String getErrorMessage(String responseBody) {
+    try {
+      JsonObject jo = gson.fromJson(responseBody, JsonObject.class);
+      return jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString();
+    } catch (Exception e) {
+      return e.getMessage();
+    }
+  }
+
+  public ServiceNowBatchRequest getPayloadRequest(List<RestRequest> restRequests) {
```
rmstar: why is this public?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,267 @@
...
+    ServiceNowBatchRequest payloadRequest = new ServiceNowBatchRequest();
+    payloadRequest.setBatchRequestId(batchRequestIdCounter.toString());
```
rmstar: Does batch request id have to be unique? Instead of a static int (potential thread safety issues) why not use a uuid?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,267 @@
...
+      restRequest.setMethod(ServiceNowConstants.HTTP_PUT);
+    }
+    restRequest.setBody(encodedData);
+    return restRequest;
+  }
+
+  /**
+   * Inserts/Updates the list of records into ServiceNow table
+   *
+   * @param restRequestList The list of restRequests
```
rmstar: It's not clear what is the difference between restRequestList and records?

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+ *  Transforms structured record to JSON object
+ */
+public class ServiceNowTransformer {
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceNowTransformer.class);
+  private static final int DEFAULT_SCALE = 2;
+  private static final String DATE_PATTERN = "yyyy-MM-dd";
+  private static final String DATE_TIME_PATTERN = "yyyy-MM-dd'T'HH:mm:ss";
+
+  @Nullable
+  public JsonObject transform(@Nullable StructuredRecord record) {
```
rmstar: when can record be null?

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+        case TIME_MILLIS:
+          LocalTime time = LocalTime.ofNanoOfDay(TimeUnit.MILLISECONDS.toNanos(((Integer) value)));
+          return time.toString();
+        case TIME_MICROS:
+          LocalTime localTime = LocalTime.ofNanoOfDay(TimeUnit.MICROSECONDS.toNanos((Long) value));
+          return localTime.toString();
+        case DATE:
+          SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_PATTERN);
+          Date date = new Date(Integer.parseInt(value.toString()) * TimeUnit.DAYS.toMillis(1));
+          return dateFormat.format(date);
```
rmstar, line 86: Please check if the data type mapping here in sync with the documentation (ServiceNow-batchsink.md).
For example, cdap date maps to glide_date (what is that?). Here we're mapping date to string.

pom.xml
sgarg-CS: Removed `<cdap.data.pipeline.version>` as it wasn't required. a1691d4500f8670c38f28b341322dac56d37f279

pom.xml
sgarg-CS: `DATETIME` data type is not available in cdap-api-common 6.1.1.
So, I had to use cdap.version 6.6.0 a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Done a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Removed the variable : `accessToken`  returning `getAccessToken` directly a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: We added a catch block in the `executePost()` method to catch `IOException` and are throwing it back to the calling method.
The IOException is being retried in `ServiceNowSinkAPIRequestImpl.createPostRequest()` method.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Yes. The same had been implemented for the ServiceNow Batch Source plugin where the schema was constructed using the first record fetched and no information was available about columns data types. https://github.com/data-integrations/servicenow-plugins/blob/d8ad62504fba128ca9d56873dbf4ab921b03652a/src/main/java/io/cdap/plugin/servicenow/source/apiclient/ServiceNowTableAPIClientImpl.java#L200

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Refactored the code. Removed the condition. Changed the return type to `void`. For codes other than 200, we're already throwing `RetryableException()` a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Changed to `private` a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Checked with `uuid` and the ServiceNow Batch API is working fine.
Replaced `batchRequestIdCounter.toString()` with `UUID.randomUUID().toString()` a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Both are the similar actually but differ when the method: `createPostRequest()` gets called from `retryUnservicedRequests()`. I had to use two different variables because while implementing the retry logic for calling the API for unserviced requests, I needed the list of all the rest requests in a single batch as well as the list of unserviced requests to pass as arguments in the method: `retryUnservicedRequests()`

This method: `retryUnservicedRequests()` calls the method `createPostRequest()` every time there is a batch of requests which are left unserviced.
It takes three arguments:

1. `restRequestList`, The total number of requests in that batch
2. `records`, The total number of unserviced requests which would be retried and,
3. `accessToken`, The access token

The `restRequestList` and `records` have the same number of requests only for the first time it gets called from `createPostRequestRetryableMode()` method.

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: Removed `@Nullable` as it cannot be null. a1691d4500f8670c38f28b341322dac56d37f279

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: For date values, ServiceNow internally converts the string to `glide_date` (an internal type used by ServiceNow system for date fields)
Same is the case with `glide_date_time` as well.

pom.xml
rmstar: I think datetime support was added in 6.4.0. Generally prefer to depend on the lowest CDAP API version that's supported so that the plugin can be compatible with as many CDAP versions are possible.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -65,17 +67,20 @@ public RestAPIResponse executeGet(RestAPIRequest request) {
...
     httpPost.setEntity(request.getEntity());
-    RestAPIResponse apiResponse = null;
+    RestAPIResponse apiResponse;
 
     try (CloseableHttpClient httpClient = HttpClientBuilder.create().build()) {
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpPost)) {
         apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
       }
+    } catch (IOException e) {
+      throw e;
```
rmstar, line 84: why are handling IOException differently from other exception types?

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
```diff
@@ -0,0 +1,143 @@
...
+
+  public JsonObject transform(StructuredRecord record) {
+    List<Schema.Field> fields = Objects.requireNonNull(record.getSchema().getFields(),
+                                                       "Schema fields cannot be empty");
+    JsonObject jsonObject = new JsonObject();
+    for (Schema.Field field : fields) {
+      String fieldName = field.getName();
+      Schema nonNullableSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable()
+        : field.getSchema();
+      jsonObject.addProperty(field.getName(), String.valueOf(extractValue(fieldName, record.get(field.getName()),
```
rmstar, line 57: Does this work correctly for decimal type? Please test all supported CDAP types.

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+/**
+ *  Transforms structured record to JSON object
+ */
+public class ServiceNowTransformer {
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceNowTransformer.class);
+  private static final int DEFAULT_SCALE = 2;
```
rmstar: why do we use scale = 2?

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
```diff
@@ -0,0 +1,143 @@
...
+    switch (fieldType) {
+      case DOUBLE:
+      case FLOAT:
+        return new BigDecimal(String.valueOf(value)).setScale(DEFAULT_SCALE, RoundingMode.HALF_UP);
+      case BOOLEAN:
+        return value;
+      case INT:
+        return value;
+      case LONG:
+        return (int) (long) value;
```
rmstar: why not just cast to long?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,256 @@
...
+            String encodedResponseBody = servicedRequestsArray.get(i).getAsJsonObject().get(ServiceNowConstants.BODY)
+              .getAsString();
+            String decodedResponseBody = new String(Base64.getDecoder().decode(encodedResponseBody));
+            String errorDetail = jsonParser.parse(decodedResponseBody).getAsJsonObject().get(ServiceNowConstants.ERROR)
+              .getAsJsonObject().get(ServiceNowConstants.ERROR_DETAIL).getAsString();
+            if (errorDetail.equals(ServiceNowConstants.ACL_EXCEPTION)) {
+              throw new IllegalStateException(String.format("Permission denied for '%s' operation.",
+                config.getOperation()));
+            } else {
+              LOG.warn("Error Response : {} ", decodedResponseBody);
```
rmstar, line 145: why aren't we throwing an exception here?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: hmm, I still don't understand the need for 2 request lists. Why can't we do something like this:
```
public void createPostRequest(List<RestRequest> restRequestList, String accessToken) {

    make POST request with ids in restRequestList
    if (response contains unserviced requests) {
       createPostRequest(unservicedRequests, accessToken);
}
```

pom.xml
sgarg-CS: downgraded from `6.6.0` to `6.4.0` 326fadb444e9fa01ddc710c3ae549590703f264e

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
sgarg-CS: We're retrying all transport exceptions while executing the HTTP POST method and the generic transport exceptions in HttpClient are represented by the standard Java java.io.IOException class
https://hc.apache.org/httpclient-legacy/exception-handling.html

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: We found an issue during testing. Have corrected the code for DECIMAL datatype.  326fadb444e9fa01ddc710c3ae549590703f264e

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: It was observed that ServiceNow Table Rest API supports up to 2 digits only after the decimal for decimal types. Anything greater than that gets rounded off.

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: Casted to just `long` 326fadb444e9fa01ddc710c3ae549590703f264e

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: ServiceNow Table Rest API gives 403 Permission Denied error with the following error message: _Error during insert of <table_name>_  if we try to insert a duplicate record and the unique constraint is applied on the column(s) of ServiceNow tables. This case falls under the `else` condition.

Sample Error message
```
{
  "error": {
    "message": "Operation Failed",
    "detail": "Error during insert of clm_condition_checker (EEf5h347_1)"
  },
  "status": "failure"
}
```

We just want to give the warning to the user with the error message and not fail the pipeline, so we aren't throwing any exception.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: We need two variables because the unserviced request array just contains the request ids. These ids are sorted and used as an index while fetching records from the original batch. 

I've separated out the two methods 326fadb444e9fa01ddc710c3ae549590703f264e:

1. `processUnservicedRequestsArray()` : Converts the unserviced request array into unserviced requests list.
2. `createPostRequest()` : Calls the ServiceNow Batch API again with the batch of unserviced requests.

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: Just FYI... We have added handling for individual datatypes like `glide_date` and `glide_date_time` as the earlier code supported only those values that were in string format.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -295,9 +317,54 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
+
+    RestAPIResponse apiResponse;
+    try {
+      String accessToken = getAccessToken();
+      requestBuilder.setAuthHeader(accessToken);
+      apiResponse = executeGet(requestBuilder.build());
+      if (!apiResponse.isSuccess()) {
+        LOG.error("Error - {}", getErrorMessage(apiResponse.getResponseBody()));
+        collector.addFailure("Unable to fetch schema for table " + tableName, null).
+          withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAME);
```
rmstar, line 346: return here as we don't want to parse the response body if there's an error

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: Why don't we want to fail the pipeline in this case?

If there's a specific case that we don't want to fail on, then make sure to check for the exact error code/message so that we don't end up ignoring unrelated errors

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,259 @@
...
+      // this condition will run when jsonElement is not null and operation is update
+    } else if (jsonElement != null) {
+      sysId = jsonElement.getAsString();
+    }
+    String data = jsonObject.toString();
+    String encodedData = Base64.getEncoder().encodeToString(data.getBytes(StandardCharsets.UTF_8));
+
+    RestRequest restRequest = new RestRequest();
+    restRequest.setUrl(getUrl(sysId));
+    restRequest.setId(counter.toString());
```
rmstar: this is not thread safe - can end up with a duplicate id if this is called concurrently from multiple threads. 
If you just need a unique id, then you can use a uuid.

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+      // Return 'null' value as it is
+      return null;
+    }
+
+    Schema.LogicalType fieldLogicalType = schema.getLogicalType();
+    // Get values of logical types properly
+    if (fieldLogicalType != null) {
+      switch (fieldLogicalType) {
+        case TIMESTAMP_MILLIS:
+          return extractUTCDateTime(TimeUnit.MILLISECONDS.toMicros((long) value)).toString();
```
rmstar: do we need to convert timestamp and time types to string here? Caller is calling String.valueOf()

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: records should be renamed as unservicedRequests

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,259 @@
...
+   *
+   * @param restRequestList         The list of rest Requests
+   * @param unservicedRequestsArray An array of unserviced requests
+   */
+  private List<RestRequest> processUnservicedRequestsArray(List<RestRequest> restRequestList,
+                                                           JsonArray unservicedRequestsArray)
+    throws InterruptedException, OAuthProblemException {
+    List<RestRequest> unservicedRequests = new ArrayList<>();
+    List<Integer> unservicedRequestsIds = new ArrayList();
+    for (int i = 0; i < unservicedRequestsArray.size(); i++) {
```
rmstar, line 201: create a Map request id-> RestRequest and lookup ids in unservicedRequestsArray in the map.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Making use of `AtomicInteger` now, for Request ID. bc67f6e1b45ebd37026d24f7b64a2d8d43fbbc88

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: added a return statement to return `null`, if there's an error. bc67f6e1b45ebd37026d24f7b64a2d8d43fbbc88

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: This error: `Error during the insert of clm_condition_checker(EEf5h347_1)` comes when a duplicate record is attempted for insertion on a column with a unique constraint. 
Assuming an average dataset having millions of records contains only a few duplicate records, we think it would be better to warn the user with a logger than to completely fail the pipeline.

If we fail the pipeline, the unique ones would anyway get inserted until the pipeline comes to a stop and there is no rollback.

I've replaced the `else` condition with `else if` to match the exact error detail message. bc67f6e1b45ebd37026d24f7b64a2d8d43fbbc88

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: We're calling `.toString()` at the end. Would it not convert the types to string?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: refactored the code by replacing the list with a map. bc67f6e1b45ebd37026d24f7b64a2d8d43fbbc88

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,272 @@
...
+   *
+   * @param restRequestsMap         The map of rest requests
+   * @param unservicedRequestsArray An array of unserviced requests
+   * @return unservicedRequestsMap The map of unserviced requests
+   */
+  private Map<String, RestRequest> processUnservicedRequestsArray(Map<String, RestRequest> restRequestsMap,
+                                                                  JsonArray unservicedRequestsArray) {
+    Map<String, RestRequest> unservicedRequestsMap = new HashMap<>();
+    for (int i = 0; i < unservicedRequestsArray.size(); i++) {
+      unservicedRequestsMap.put(unservicedRequestsArray.get(i).getAsString(),
```
rmstar: Store the request id in a variable, log an error an ignore request id if it's not found in restRequestsMap
something like this - 
```java
String requestId = unservicedRequestsArray.get(i).getAsString();
RestRequest request = restRequestsMap.get(requestId);
if (request == null) {
   // log error and continue
}
unservicedRequestsMap.put(requestId, request);
```

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: I see. What happens if some records in the batch are duplicates, others are not? Is this error returned for duplicate records, unique records are inserted successfully? Can you please verify what happens in this case?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,272 @@
...
+
+  /**
+   * Retries to insert/update the records into ServiceNow table when RetryableException is thrown          .
+   *
+   * @param restRequestsMap The map of rest Requests
+   */
+  public void createPostRequestRetryableMode(Map<String, RestRequest> restRequestsMap) throws ExecutionException,
+    RetryException {
+    // Reset request counter
+    atomicCounter.set(1);
```
rmstar: why do we need to reset the counter? Any reason why we need incrementing ids instead of just using a UUID like you did in ServiceNowBatchRequest.setBatchRequestId()?

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
rmstar: Calling .toString() here is not necessary because the caller is converting to String
```java
String.valueOf(extractValue(fieldName, record.get(field.getName())
```

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Yes, this error is returned when duplicated records are attempted for insertion on a column having a unique constraint.

In the case when some records are duplicate and others are unique in a batch, then the unique ones are inserted successfully and duplicate ones are not inserted and an error message is thrown. The error message contains the name of the table and the duplicate value. This error is logged and displayed as a warning to the user. The pipeline continues to run normally as usual.

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: Removed `.toString()` 0a1c20a052aca8be095d4e1cde499b71fed53861

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: The counter was being reset to re-initialize the starting request in a batch to 1.  We're using incrementing ids when we had `restRequestsList`, as it was easier to process the unserviced requests. 

Since we are using `restRequestsMap` now, the incrementing ids have been replaced by UUID. 0a1c20a052aca8be095d4e1cde499b71fed53861

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: done. 0a1c20a052aca8be095d4e1cde499b71fed53861

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,262 @@
...
+   * @return unservicedRequestsMap The map of unserviced requests
+   */
+  private Map<String, RestRequest> processUnservicedRequestsArray(Map<String, RestRequest> restRequestsMap,
+                                                                  JsonArray unservicedRequestsArray) {
+    Map<String, RestRequest> unservicedRequestsMap = new HashMap<>();
+    for (int i = 0; i < unservicedRequestsArray.size(); i++) {
+      String requestId = unservicedRequestsArray.get(i).getAsString();
+      RestRequest request = restRequestsMap.get(requestId);
+      if (request == null) {
+        LOG.error("No Rest Request found for Request Id {}", requestId);
```
rmstar, line 205: you should continue here, otherwise you'll add a null value

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,262 @@
...
+
+  private ServiceNowBatchRequest getPayloadRequest(Map<String, RestRequest> restRequests) {
+    ServiceNowBatchRequest payloadRequest = new ServiceNowBatchRequest();
+    payloadRequest.setBatchRequestId(UUID.randomUUID().toString());
+    payloadRequest.setRestRequests(new ArrayList<>(restRequests.values()));
+    return payloadRequest;
+  }
+
+  /**
+   * Process the unserviced requests array into an unserviced requests map.    .
```
rmstar: nit remove trailing '    .'

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,262 @@
...
+   *
+   * @param restRequestsMap         The map of rest requests
+   * @param unservicedRequestsArray An array of unserviced requests
```
rmstar: array on unserviced request ids

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,262 @@
...
+   */
+  public void createPostRequestRetryableMode(Map<String, RestRequest> restRequestsMap) throws ExecutionException,
+    RetryException {
+    String accessToken = restApi.getAccessTokenRetryableMode();
+    Callable<Boolean> fetchRecords = () -> {
+      createPostRequest(restRequestsMap, accessToken);
+      return true;
+    };
+
+    Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
```
rmstar: if you're not doing anything with the return value you can just make it a Retryer<Void>

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: I see, thanks. When this happens, is `unserviced_requests` set in the response JSON? i.e. just want to make sure an exception is not thrown in line 154
```
        JsonArray unservicedRequestsArray = responseJSON.get(ServiceNowConstants.UNSERVICED_REQUESTS).getAsJsonArray();
```

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,262 @@
...
+              .getAsJsonObject().get(ServiceNowConstants.ERROR_DETAIL).getAsString();
+            
+            if (errorDetail.equals(ServiceNowConstants.ACL_EXCEPTION)) {
+              throw new IllegalStateException(String.format("Permission denied for '%s' operation.",
+                                                            config.getOperation()));
+            } else if (errorDetail.contains(ServiceNowConstants.INSERTION_ERROR)) {
+              LOG.warn("Error Response : {} ", decodedResponseBody);
+            } else if (errorDetail.contains((ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED))) {
+              failedRequestId = servicedRequestsArray.get(i).getAsJsonObject().get(ServiceNowConstants.ID);
+            }
```
rmstar, line 150: Don't you need another else clause to handle any other exceptions that might happen and throw an exception in that case?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,262 @@
...
+                                                                                          unservicedRequestsArray);
+          // Retry unserviced requests
+          createPostRequest(unservicedRequestsMap, accessToken);
+        }
+      }
+    } catch (IOException e) {
+      LOG.error("Error while connecting to ServiceNow", e.getMessage());
+      throw new RetryableException();
+    } catch (Exception e) {
+      throw e;
```
rmstar: I don't think you need to catch the exception here since you're just throwing it

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowRecordWriter.java
```diff
@@ -0,0 +1,73 @@
...
+  public void write(NullWritable key, JsonObject jsonObject) throws IOException {
+
+    RestRequest restRequest = servicenowSinkAPIImpl.getRestRequest(jsonObject);
+    restRequestsMap.put(restRequest.getId(), restRequest);
+    if (restRequestsMap.size() == ServiceNowConstants.RECORDS_PER_BATCH) {
+      try {
+        servicenowSinkAPIImpl.createPostRequestRetryableMode(restRequestsMap);
+      } catch (RetryException | ExecutionException exception) {
+        restRequestsMap.clear();
+        throw new IOException(exception.getCause());
```
rmstar: throw new IOException("Error writing to ServiceNow", exception);

Similar thing in close()

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
```diff
@@ -28,9 +28,19 @@
  */
 public class ServiceNowTableAPIRequestBuilder extends RestAPIRequest.Builder {
   private static final String TABLE_API_URL_TEMPLATE = "%s/api/now/table/%s";
+  private static final String BATCH_API_URL_TEMPLATE = "%s/api/now/v1/batch";
+  private static final String SCHEMA_API_URL_TEMPLATE = "%s/api/now/doc/table/schema/%s";
```
rmstar, line 44: What's the difference between these 3 APIs, and when should each one be used?
  private static final String TABLE_API_URL_TEMPLATE = "%s/api/now/table/%s";
  private static final String BATCH_API_URL_TEMPLATE = "%s/api/now/v1/batch";
  private static final String SCHEMA_API_URL_TEMPLATE = "%s/api/now/doc/table/schema/%s";

Can you add some comments to clarify?

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
rmstar: Please add this as a comment in the code

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Yes, it is always returned in the Response Payload.

This is what the response `JSON` looks like.
image

If there are no unserviced requests, then the JSON array is empty.
image

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowTransformer.java
sgarg-CS: added. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: yep, it's unnecessary. Removed it. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/transform/ServiceNowRecordWriter.java
sgarg-CS: done at both places. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
sgarg-CS: Added comments in the code for all three URLs for better understanding. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Replaced `Retryer<Boolean>` with `Retryer` e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Updated the comment. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: removed. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: done. e8b33614f910fa16f06ffcf185ed0d3a23d9fda1

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Added an `else` condition throwing exception to handle all other errors returned by ServiceNow Rest API.  383670aa689fe2c05e336d3da0fc2b727f94a760

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
rmstar: annotate as Nullable. Also, the caller needs to handle null schema.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -0,0 +1,263 @@
...
+        LOG.error("Error - {}", getErrorMessage(apiResponse.getResponseBody()));
+        throw new RetryableException();
+      } else {
+        JsonObject responseJSON = jsonParser.parse(apiResponse.getResponseBody()).getAsJsonObject();
+        JsonArray servicedRequestsArray = responseJSON.get(ServiceNowConstants.SERVICED_REQUESTS).getAsJsonArray();
+        JsonElement failedRequestId = null;
+        for (int i = 0; i < servicedRequestsArray.size(); i++) {
+          int statusCode = servicedRequestsArray.get(i).getAsJsonObject().get(ServiceNowConstants.STATUS_CODE)
+            .getAsInt();
+          if (statusCode == HttpStatus.SC_FORBIDDEN || statusCode == HttpStatus.SC_BAD_REQUEST ||
```
rmstar: do we need to handle other non-200 status codes?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: Thanks for clarifying. We are throwing IllegalStateException when there's an error in processing one of the requests - this doesn't get retried. Is that the correct behavior?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Added a case when ServiceNow Table API returns HTTP status code: 404 if _sys_id_ present in the request payload is not found in ServiceNow. (For update operation)


The following error message is returned by the API in this case : 
```
{
  "error": {
    "message": "No Record found",
    "detail": "Record doesn't exist or ACL restricts the record retrieval"
  },
  "status": "failure"
}
```

141d14bc854fc0889d5a6fabae03e8127abe5c59

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: `IllegalStateException` is thrown in the following 3 cases when

1. No `sys_id` is found in the record to be updated.
2. This error message comes in response payload: **ACL Exception Insert Failed due to security constraints** (seen in duplicate records)
3. All other error messages returned by ServiceNow Table API apart from the ones which have been handled in the code.

All the above cases do not need to be retried.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Annotated with `@Nullable` 141d14bc854fc0889d5a6fabae03e8127abe5c59

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: Thanks for clarifying!

src/main/java/io/cdap/plugin/servicenow/sink/ServiceNowSinkConfig.java
```diff
@@ -175,21 +135,18 @@ public void validateSchema(Schema schema, FailureCollector collector) {
       || containsMacro(ServiceNowConstants.PROPERTY_OPERATION) || containsMacro(PROPERTY_EXTERNAL_ID_FIELD)) {
       return;
     }
-
-    Schema tableSchema = getTableMetaData(tableName, this).getSchema();
+    ServiceNowTableAPIClientImpl restApi = new ServiceNowTableAPIClientImpl(this);
+    Schema tableSchema = restApi.fetchServiceNowTableSchema(tableName, collector);
```
rmstar, line 139: need to check if tableSchema is null here

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
rmstar: Instead of checking specific error codes it would be better to handle all 4xx and 5xx codes as errors. So in the future, if the ServiceNow API returns a different error code like 503 (Service Unavailable), this code should still work as expected.
For example, see https://github.com/cdapio/cdap/blob/develop/cdap-runtime-ext-dataproc/src/main/java/io/cdap/cdap/runtime/spi/provisioner/dataproc/DataprocClient.java#L870

src/main/java/io/cdap/plugin/servicenow/sink/ServiceNowSinkConfig.java
sgarg-CS: Added null check. f453a61bee4a0cffcfd69509adb54fa1b7933f27

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Refactored the **if** condition to check for all 4xx as well as 5xx HTTP Status Codes f453a61bee4a0cffcfd69509adb54fa1b7933f27

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
rmstar: Can you add this as a comment in the code.

Edit: github UI is not showing the previous comments in this thread for some reason. So providing the context here in case it's not clear which comment I'm talking about :) - https://github.com/data-integrations/servicenow-plugins/pull/34/files#r958614835

rmstar (approved):
👍 please squash commits after addressing minor comment

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
sgarg-CS: Added the same as a comment in the code. f372a0337e75290ed56764e7d12d509539242fa0

sgarg-CS:
> 👍 please squash commits after addressing minor comment

All commits have been squashed. Thank you.

PR #52, opened 2023-03-24
Connection management e2e tests

Vipinofficial11:


Vipinofficial11:
Moving the PR in Draft for now as working on e2e of Connection Management functionality for other plugins as well and noticing some steps which can be generic. ill raise a PR first in the cdap-e2e-test framework.

PR #60, opened 2023-06-27
ServiceNow Batch Source GA Release changes.

sgarg-CS:
We're doing a GA release for ServiceNow Batch Source Plugin. This PR addresses the following issues:

- [PLUGIN-1650](https://cdap.atlassian.net/browse/PLUGIN-1650)
- [PLUGIN-1662](https://cdap.atlassian.net/browse/PLUGIN-1662)

Please refer to the code change doc [here](https://docs.google.com/document/d/1eejV4ExEPxcxC8TnW2Q1HNWXH03ZoxvwmqlkI2HT90I/edit)

[PLUGIN-1650]: https://cdap.atlassian.net/browse/PLUGIN-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLUGIN-1652]: https://cdap.atlassian.net/browse/PLUGIN-1652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLUGIN-1662]: https://cdap.atlassian.net/browse/PLUGIN-1662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
```diff
@@ -204,8 +203,9 @@ private List<StructuredRecord> getTableData(String tableName, int limit) throws
         StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
         for (Schema.Field field : tableFields) {
           String fieldName = field.getName();
-          Object fieldValue = config.convertToValue(fieldName, field.getSchema(), result.get(i));
-          recordBuilder.set(fieldName, fieldValue);
+          //Object fieldValue = config.convertToValue(fieldName, field.getSchema(), result.get(i), recordBuilder);
+          config.convertToValue(fieldName, field.getSchema(), result.get(i), recordBuilder);
+          //recordBuilder.set(fieldName, fieldValue);
```
Aryan-Verma: Please remove the old code

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -93,25 +88,20 @@ static Set<ServiceNowTableInfo> fetchTablesInfo(ServiceNowConnectorConfig conf,
...
+  private static ServiceNowTableInfo getTableMetaData(String tableName, ServiceNowConnectorConfig conf) {
     // Call API to fetch first record from the table
     ServiceNowTableAPIClientImpl restApi = new ServiceNowTableAPIClientImpl(conf);
 
-    ServiceNowTableDataResponse response = restApi.fetchTableSchema(tableName, valueType, startDate, endDate,
-                                                                    true);
-    if (response == null) {
-      return null;
+    Schema schema;
+    int recordCount;
```
Aryan-Verma, line 96:
```suggestion
    Schema schema = null;
    int recordCount = 0;
```

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -134,7 +124,7 @@ public List<InputSplit> getSplits(JobContext jobContext) throws IOException, Int
       String tableName = tableInfo.getTableName();
       int totalRecords = tableInfo.getRecordCount();
 
-      int pages = (tableInfo.getRecordCount() / ServiceNowConstants.PAGE_SIZE);
+      int pages = (totalRecords / ServiceNowConstants.PAGE_SIZE);
       if (tableInfo.getRecordCount() % ServiceNowConstants.PAGE_SIZE > 0) {
```
Aryan-Verma:
```suggestion
      if (totalRecords % ServiceNowConstants.PAGE_SIZE > 0) {
```

src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
```diff
@@ -77,7 +77,7 @@ private Schema createSchema(ServiceNowColumn column) {
       case "glide_date_time":
         return Schema.of(Schema.LogicalType.DATETIME);
       case "glide_time":
-        return Schema.of(Schema.LogicalType.TIME_MICROS);
+        return Schema.of(Schema.LogicalType.TIME_MILLIS);
```
Aryan-Verma: Why are we changing it to TIME_MILLIS?

widgets/ServiceNow-batchsource.json
```diff
@@ -128,6 +128,14 @@
     {
       "label": "Table Mode",
       "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Table Name",
+          "name": "tableName",
+          "widget-attributes": {
+            "placeholder": "ServiceNow table name from which data to be fetched"
```
Aryan-Verma: nit:
```suggestion
            "placeholder": "ServiceNow table name to fetch data from"
```

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
sgarg-CS: removed. 802f6bfaed7ac76ee29c5734bdd0d572849d78ae

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
sgarg-CS: IntelliJ shows these initializations as redundant.

image

image

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
sgarg-CS: changed 802f6bfaed7ac76ee29c5734bdd0d572849d78ae

src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
sgarg-CS: `TIME_MILLIS` was found to be a more suitable data type for `glide_time` during testing compared to `TIME_MICROS`

widgets/ServiceNow-batchsource.json
sgarg-CS: updated 802f6bfaed7ac76ee29c5734bdd0d572849d78ae

DJSagarAhire:
Can you please add a description to the PR that describes what the PR is doing (and why) so that it is easier to review? Also add JIRA issues that the PR addresses in title if applicable.

sgarg-CS:
> Can you please add a description to the PR that describes what the PR is doing (and why) so that it is easier to review? Also add JIRA issues that the PR addresses in title if applicable.

Added the description and the JIRA issues fixed in this PR.
Please refer to the code change doc [here](https://docs.google.com/document/d/1eejV4ExEPxcxC8TnW2Q1HNWXH03ZoxvwmqlkI2HT90I/edit)

albertshau (commented):
please open one PR per issue so it is clear what problem the code is fixing.

Changing data types is also a backwards incompatible change. What we have done in other cases is to support reading the field as both the old type and the new type. That way, when a pipeline is upgraded, it will continue working as it did before instead of breaking.

sgarg-CS:
> please open one PR per issue so it is clear what problem the code is fixing.
> 
> Changing data types is also a backwards incompatible change. What we have done in other cases is to support reading the field as both the old type and the new type. That way, when a pipeline is upgraded, it will continue working as it did before instead of breaking.

Created a smaller [PR](https://github.com/data-integrations/servicenow-plugins/pull/62) for two of the issues.

sgarg-CS:
> please open one PR per issue so it is clear what problem the code is fixing.
> 
> Changing data types is also a backwards incompatible change. What we have done in other cases is to support reading the field as both the old type and the new type. That way, when a pipeline is upgraded, it will continue working as it did before instead of breaking.

This feature was not available previously. All the datatypes were getting mapped to `String` earlier. We introduced a new API call to ServiceNow to fetch the schema to map the datatypes correctly. 
The ServiceNow Batch Source plugin was in preview mode. It will be going for GA release for the first time, so the backward incompatibility shouldn't cause an issue.

docs/ServiceNow-batchsource.md
```diff
@@ -74,8 +74,8 @@ Data Types Mapping
     | boolean                        | boolean               |                                                    |
     | reference                      | string                |                                                    |
     | currency                       | string                |                                                    |
-    | glide_date                     | string                |                                                    |
-    | glide_date_time                | string                |                                                    |
+    | glide_date                     | date                  |                                                    |
+    | glide_date_time                | date_time             |                                                    |
```
albertshau: date_time -> datetime, to be consistent with how it shows up in other places

albertshau: The Comment column is empty for everything and should be removed.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -326,9 +187,12 @@ public List<Map<String, Object>> parseResponseToResultListOfMap(String responseB
   private String getErrorMessage(String responseBody) {
     try {
       JsonObject jo = gson.fromJson(responseBody, JsonObject.class);
-      return jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString();
+      return String.format("%s:%s",
+                           jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString(),
```
albertshau: can any of these fields in the json object ever be null?

Even if not, it is safer to do a null check anyway to be more defensive. For example, maybe the ServiceNow API changes at some point in the future.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
    */
   @Nullable
-  public Schema fetchServiceNowTableSchema(String tableName, FailureCollector collector) {
+  public Schema fetchTableSchema(String tableName, FailureCollector collector) {
+    Schema schema = null;
+    try {
+      schema = fetchTableSchema(tableName);
+    } catch (OAuthProblemException | OAuthSystemException | RuntimeException e) {
+      LOG.error("Error in connection - {}", e);
+      collector.addFailure("Connection failed. Unable to fetch schema for table " + tableName, null);
```
albertshau: this hides the actual cause of the error from the user. The exception message should be used somewhere, otherwise there will be no way to tell what went wrong.

albertshau: should also add the stacktrace to the failure using `withStacktrace`, it will normally be useful.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
+    }
+    return schema;
+  }
+
+  @VisibleForTesting
+  public List<SchemaResponse> parseSchemaResponse(String responseBody) {
```
albertshau: shouldn't have parsing methods like this. It is more error prone and less clear than using classes that represent the entire response. For example:

```
public class SchemaResponse {
  private final List<SchemaField> result;
}
```

where the current SchemaResponse class is renamed to SchemaField. Then deserialize the whole thing directly: 

```
SchemaResponse response = gson.fromJson(apiResponse.getResponseBody(), SchemaResponse.class);
```

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -68,8 +67,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   private static final String FIELD_UPDATED_ON = "sys_updated_on";
   private static final String OAUTH_URL_TEMPLATE = "%s/oauth_token.do";
   private static final Gson gson = new Gson();
```
albertshau: static final variables should be all caps -- GSON

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       this.conf.getRestApiEndpoint(), tableName, true)
       .setExcludeReferenceLink(true);
 
     RestAPIResponse apiResponse;
+    String accessToken = getAccessToken();
+    requestBuilder.setAuthHeader(accessToken);
+    apiResponse = executeGet(requestBuilder.build());
+    if (!apiResponse.isSuccess()) {
+      LOG.error("Error - {}", getErrorMessage(apiResponse.getResponseBody()));
```
albertshau: log and throw is an anti-pattern as it results in excessive logging and potentially error logs for non-errors (like if the caller decides to retry which succeeds). This log message should be removed.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
+      throw new RuntimeException("Error - " + getErrorMessage(apiResponse.getResponseBody()));
+    }
+    List<SchemaResponse> result = parseSchemaResponse(apiResponse.getResponseBody());
+    List<ServiceNowColumn> columns = new ArrayList<>();
+
+    if (result != null && !result.isEmpty()) {
```
albertshau: if it's null or empty, it looks like it will result in an exception when the schema is created. That will be a more confusing error message than if it is handled directly here instead:

```
if (result == null || result.isEmpty()) {
  // throw some exception with useful message
}
for (SchemaResponse field : result) {
  ....
}
```

albertshau: Also, SchemaResponse is an odd name for a field. It would be more descriptive to name it something like SchemaField and have SchemaResponse represent the entire response body.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
-      }
-      List<SchemaResponse> result = parseSchemaResponse(apiResponse.getResponseBody());
-      List<ServiceNowColumn> columns = new ArrayList<>();
-
-      if (result != null && !result.isEmpty()) {
-        for (SchemaResponse field : result) {
-          columns.add(new ServiceNowColumn(field.getName(), field.getInternalType()));
-        }
+      } else {
+        LOG.info(apiResponse.getResponseBody());
```
albertshau: should not be logging response bodies. Customer data should never be in the logs.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
       }
-      return SchemaBuilder.constructSchema(tableName, columns);
-    } catch (OAuthSystemException | OAuthProblemException e) {
-      LOG.error("Error in connection - {}", e);
-      collector.addFailure("Connection failed. Unable to fetch schema for table " + tableName, null);
+    } catch (OAuthSystemException | OAuthProblemException | UnsupportedEncodingException e) {
+      LOG.error("Error in creating a new record", e);
```
albertshau: similar comment, don't log and rethrow

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
 
-  @VisibleForTesting
-  public List<SchemaResponse> parseSchemaResponse(String responseBody) {
-    JsonObject jo = gson.fromJson(responseBody, JsonObject.class);
-    JsonArray ja = jo.getAsJsonArray(ServiceNowConstants.RESULT);
-    Type type = new TypeToken<List<SchemaResponse>>() {
-    }.getType();
-    return gson.fromJson(ja, type);
+  private String getSystemId(RestAPIResponse apiResponse) {
+    JsonObject jsonObject = gson.fromJson(apiResponse.getResponseBody(), JsonObject.class);
```
albertshau: similar comment as above, we shouldn't be doing any json parsing or working directly with JsonObject. There should be a concrete class that represents the response body.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
+      } else {
+        LOG.info(apiResponse.getResponseBody());
       }
-      return SchemaBuilder.constructSchema(tableName, columns);
-    } catch (OAuthSystemException | OAuthProblemException e) {
-      LOG.error("Error in connection - {}", e);
-      collector.addFailure("Connection failed. Unable to fetch schema for table " + tableName, null);
+    } catch (OAuthSystemException | OAuthProblemException | UnsupportedEncodingException e) {
+      LOG.error("Error in creating a new record", e);
+      throw new RuntimeException("Error in creating a new record");
```
albertshau: when wrapping exceptions, the cause must always be specified, otherwise the call stack is lost.

```
throw new RuntimeException(..., e);
```

that being said, it's normally not a good idea to rethrow checked exceptions as runtime exceptions in code that we completely control. We should either be propagating the exception up instead of catching and rethrowing (normally better), or we should wrap it in an IOException instead of RuntimeException.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -375,45 +238,146 @@ public List<Map<String, Object>> fetchTableRecordsRetryableMode(String tableName
...
+
+    ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
+      this.conf.getRestApiEndpoint(), tableName, false)
+      .setQuery(query);
+
+    RestAPIResponse apiResponse = null;
+    String accessToken = getAccessToken();
+    requestBuilder.setAuthHeader(accessToken);
+    apiResponse = executeGet(requestBuilder.build());
+    JsonObject jsonObject = gson.fromJson(apiResponse.getResponseBody(), JsonObject.class);
```
albertshau: similar comment as elsewhere, use concrete java classes and not JsonObject

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -20,42 +20,61 @@
...
  * PluginConfig for ServiceNow Connector
  */
 public class ServiceNowConnectorConfig extends PluginConfig {
+  private static final String DATE_PATTERN = "yyyy-MM-dd";
+  private static final String DATE_TIME_PATTERN = "yyyy-MM-dd HH:mm:ss";
+  private static final String TIME_PATTERN = "HH:mm:ss";
+
   @Name(ServiceNowConstants.PROPERTY_CLIENT_ID)
   @Macro
+  @Nullable
```
albertshau, line 37: why are these all getting changed to nullable? They seem like they would be required?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -136,22 +155,62 @@ public void validateConnection(FailureCollector collector) {
     }
   }
 
-  public Object convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record) {
-    Schema.Type fieldType = fieldSchema.getType();
+  public void convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record,
+                               StructuredRecord.Builder recordBuilder) {
     Object fieldValue = record.get(fieldName);
+    if (fieldValue == null || fieldValue.toString().isEmpty()) {
```
albertshau: this `toString().isEmpty()` check is odd. When would that be empty instead of null, and when would we ever want to do this for a non-string object?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -172,17 +231,23 @@ public Double convertToDoubleValue(Object fieldValue) {
     if (fieldValue instanceof String && Strings.isNullOrEmpty(String.valueOf(fieldValue))) {
       return null;
     }
-
-    return Double.parseDouble(String.valueOf(fieldValue));
+    try {
+      return NumberFormat.getNumberInstance(java.util.Locale.US).parse(String.valueOf(fieldValue)).doubleValue();
```
albertshau: I'm confused by all these convert methods. They all do `String.valueOf(fieldValue)`. What are the actual classes in the Map? Are they always strings? If so, why is it being passed as `Map<String, Object>` instead of `Map<String, String>`? If these values are other classes, why aren't they being cast to the proper class instead of being converted to a string and then converted back?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -172,17 +231,23 @@ public Double convertToDoubleValue(Object fieldValue) {
...
+    } catch (ParseException e) {
+      return null;
```
albertshau: this seems like a bad idea. If invalid values are read as null, there is no way to differentiate between an actual null and an bad value.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -136,22 +155,62 @@ public void validateConnection(FailureCollector collector) {
...
+        return;
       case INT:
-        return convertToIntegerValue(fieldValue);
+        recordBuilder.set(fieldName, convertToIntegerValue(fieldValue));
+        return;
       case BOOLEAN:
-        return convertToBooleanValue(fieldValue);
+        recordBuilder.set(fieldName, convertToBooleanValue(fieldValue));
+        return;
       case UNION:
```
albertshau: I don't see unions in the type mapping table from the documentation. If this is only to handle nulls, nulls are already handled at the start of this method so this case should be removed.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -70,8 +83,8 @@ public StructuredRecord getCurrentValue() throws IOException {
     try {
       for (Schema.Field field : tableFields) {
         String fieldName = field.getName();
-        Object fieldValue = multiSourcePluginConf.getConnection().convertToValue(fieldName, field.getSchema(), row);
-        recordBuilder.set(fieldName, fieldValue);
+        multiSourcePluginConf.getConnection().convertToValue(fieldName, field.getSchema(), row,
+                                                                                 recordBuilder);
```
albertshau: fix indentation

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -20,42 +20,61 @@
...
  */
 public class ServiceNowConnectorConfig extends PluginConfig {
+  private static final String DATE_PATTERN = "yyyy-MM-dd";
+  private static final String DATE_TIME_PATTERN = "yyyy-MM-dd HH:mm:ss";
+  private static final String TIME_PATTERN = "HH:mm:ss";
+
   @Name(ServiceNowConstants.PROPERTY_CLIENT_ID)
   @Macro
+  @Nullable
   @Description(" The Client ID for ServiceNow Instance.")
```
albertshau: remove space in front of 'The'

docs/ServiceNow-batchsource.md
sgarg-CS: > date_time -> datetime, to be consistent with how it shows up in other places

done e63c214cf864e664ca24072f21c719c1a777b9c4

> The Comment column is empty for everything and should be removed.

Removed the Comment column e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: didn't get to see `null` for any of these fields during testing. Added null checks for all fields just in case. e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: > this hides the actual cause of the error from the user. The exception message should be used somewhere, otherwise there will be no way to tell what went wrong.

Added exception message to the logger. e63c214cf864e664ca24072f21c719c1a777b9c4

> should also add the stacktrace to the failure using `withStacktrace`, it will normally be useful.

**Do we want to add stacktrace to the failure collector?** in order to display the message with the complete stacktrace at the time of validation on the UI?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Created a new class `SchemaResponse` and renamed the existing one to `SchemaField` e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: done e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Removed the logger e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: > if it's null or empty, it looks like it will result in an exception when the schema is created. That will be a more confusing error message than if it is handled directly here instead:
> 
> ```
> if (result == null || result.isEmpty()) {
>   // throw some exception with useful message
> }
> for (SchemaResponse field : result) {
>   ....
> }
> ```
Throwing Exception with a message.

Introduced a new class `SchemaResponse` as per the previous review comment and replaced the current one by `SchemaField` e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Removed the logger e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Removed the logger. e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Changed to IOException and specified the casue. e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Client_ID is required for establishing a connection. Must have been missed out getting adding @Nullable annotation by the developer earlier. e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: fixed e63c214cf864e664ca24072f21c719c1a777b9c4

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: removed e63c214cf864e664ca24072f21c719c1a777b9c4

src/e2e-test/resources/pluginParameters.properties
```diff
@@ -45,19 +45,19 @@ UPDATE=update
 pagesize=200
 
 ##ExpectedSchemaJSONs
-schema.table.receiving.slip.line=[{"key":"cost","value":"string"},{"key":"quantity","value":"string"},\
-  {"key":"purchase_line","value":"string"},{"key":"sys_mod_count","value":"string"},\
-  {"key":"received","value":"string"},{"key":"requested_for","value":"string"},\
-  {"key":"sys_updated_on","value":"string"},{"key":"sys_tags","value":"string"},{"key":"number","value":"string"},\
+schema.table.receiving.slip.line=[{"key":"quantity","value":"int"},{"key":"cost","value":"string"}\
```
DJSagarAhire: There is no comma after `{"key":"cost","value":"string"}`. How is the test working without the comma?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
DJSagarAhire: If the Client ID is required, then the field should *not* be nullable, correct? We should be removing the `@Nullable` annotation from fields that are required.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -136,22 +155,62 @@ public void validateConnection(FailureCollector collector) {
     }
   }
 
-  public Object convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record) {
-    Schema.Type fieldType = fieldSchema.getType();
+  public void convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record,
+                               StructuredRecord.Builder recordBuilder) {
```
DJSagarAhire: Please fix the spacing here. This line is currently not aligned with anything in particular.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -136,22 +155,62 @@ public void validateConnection(FailureCollector collector) {
...
+            recordBuilder.setDateTime(fieldName, LocalDateTime.parse(fieldValue.toString(), dateTimeFormatter));
+          } catch (DateTimeParseException exception) {
+            throw new UnexpectedFormatException(
+              String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
+                            fieldName, fieldSchema.getDisplayName(), fieldValue), exception);
+          }
+          return;
+        case DATE:
+          DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern(DATE_PATTERN);
+          recordBuilder.setDate(fieldName, LocalDate.parse(fieldValue.toString(), dateFormatter));
```
DJSagarAhire: Why do we not want to handle any `DateTimeParseException` here the way we have above for `DATETIME`?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -136,22 +155,62 @@ public void validateConnection(FailureCollector collector) {
...
+          return;
+        case TIME_MILLIS:
+          DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern(TIME_PATTERN);
+          recordBuilder.setTime(fieldName, LocalTime.parse(fieldValue.toString(), timeFormatter));
```
DJSagarAhire: Why do we not want to handle any `DateTimeParseException` here the way we have above for `DATETIME`?

src/main/java/io/cdap/plugin/servicenow/sink/model/SchemaField.java
```diff
@@ -0,0 +1,59 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.servicenow.sink.model;
+
+/**
+ * Model class for Schema Field from Schema API
+ */
+public class SchemaField {
```
DJSagarAhire: The name of this class can be easily confused with `Schema.Field`. For example I see a list below called `schemaFields` which is a list of `Schema.Field` rather than `SchemaField`. If this is specific to ServiceNow, the class name should ideally indicate that.

DJSagarAhire:
> Please refer to the code change doc [here](https://docs.google.com/document/d/1eejV4ExEPxcxC8TnW2Q1HNWXH03ZoxvwmqlkI2HT90I/edit)

I've requested access to this doc yesterday as well, will do another round of review once I have the access.

DJSagarAhire:
Received access to the one-pager. It is merely repeating the names of the 2 JIRA issues in the PR description. Requesting addition of some details on the root cause of the errors in the JIRA and a short description of the design approach in fixing these.

src/e2e-test/resources/pluginParameters.properties
sgarg-CS: added a comma after `{"key":"cost","value":"string"}`. a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Fixed. a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Handled using try-catch block a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Handled using try-catch block a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/sink/model/SchemaField.java
sgarg-CS: Renamed the class to `ServiceNowSchemaField` a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: The UNION case is required for fields having nullable schema types. At the start of this method, the `null` check is for the field value, not the schema.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Yes, they are always strings. 
Removed `String.valueOf(fieldValue)` and Replaced `Map<String, Object>` with `Map<String, String>` a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: ServiceNow sometimes returns blank or empty values instead of `null`. As all values are string, therefore remove the call to `toString()` method. a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: There was a bug identified by the QA team when we were using the connection manager, the pipeline is not getting deployed with the error that all these properties are non-nullable.

For instance, this error comes when you try to deploy the pipeline without giving the pipeline name.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Added APIResponse class and deserialized the response directly to the Java object. a963c57e661447a3f1a6e970cdabc6f76964b313

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Added APIResponse class and deserialized the response directly to the Java object. a963c57e661447a3f1a6e970cdabc6f76964b313

sgarg-CS:
> Received access to the one-pager. It is merely repeating the names of the 2 JIRA issues in the PR description. Requesting addition of some details on the root cause of the errors in the JIRA and a short description of the design approach in fixing these.

Added more details to the one-pager about the JIRA issues and the approach used to fix those issues.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
albertshau: this is a confusing way to do it. Instead, after the null check earlier, do:

```
fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema
```

and remove this case statement

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -136,22 +154,74 @@ public void validateConnection(FailureCollector collector) {
     }
   }
 
-  public Object convertToValue(String fieldName, Schema fieldSchema, Map<String, Object> record) {
+  public void convertToValue(String fieldName, Schema fieldSchema, Map<String, String> record,
```
albertshau: these conversion methods are not related to the config at all and should be somewhere else. It should be moved into some sort of record converter class

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
albertshau: this is still doing valueOf() and the method is still taking an Object instead of String as an argument.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
albertshau: still have this comment, do not do this anywhere. It is wrong behavior and will lead to data loss.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
albertshau: stacktrace will include it in the response so it isn't lost.  The UI doesn't show it right now, but it is still useful to have it if somebody makes the rest call manually, or uses the browser developer tools to see the response body.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Moved the methods to a utility class : `ServiceNowRecordConverter` f58c78717ea204f30540f24a0352ce21ef7ed47b

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: removed the case statement and added the schema nullable check at the beginning. f58c78717ea204f30540f24a0352ce21ef7ed47b

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: Missed it earlier. Have addressed it now. Throwing exception to the calling method and then throwing `IllegalStateException` with a detailed message. f58c78717ea204f30540f24a0352ce21ef7ed47b

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
sgarg-CS: corrected. Replaced with `String` arg and removed `String.valueOf()` f58c78717ea204f30540f24a0352ce21ef7ed47b

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Included the stack trace in the failure message. f58c78717ea204f30540f24a0352ce21ef7ed47b

DJSagarAhire (approved):
LGTM, but ensure that all of @albertshau's comments are also addressed before merging. Also squash all the commits into a single one.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -0,0 +1,138 @@
...
+      throw new IllegalStateException(
+        String.format("Error occurred while parsing the value '%s' for field '%s'", fieldValue, fieldName));
+    }
+
+  }
+
+  @VisibleForTesting
+  public static Double convertToDoubleValue(String fieldValue) throws ParseException {
+    try {
+      return NumberFormat.getNumberInstance(java.util.Locale.US).parse(fieldValue).doubleValue();
```
albertshau: style: `Locale.US` and import java.util.Locale

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -0,0 +1,138 @@
...
+    } catch (ParseException exception) {
```
albertshau, line 116: there is no need for the try-catch here if we're just throwing as-is.

The date/time fields throw an UnexpectedFormatException, we can change these methods to do the same.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -0,0 +1,138 @@
...
+    try {
+      return NumberFormat.getNumberInstance(java.util.Locale.US).parse(fieldValue).intValue();
+    } catch (ParseException exception) {
+      throw exception;
+    }
+  }
+
+  @VisibleForTesting
+  public static Boolean convertToBooleanValue(String fieldValue) {
+    return Boolean.parseBoolean(fieldValue);
```
albertshau: this will return true if the fieldValue is `true` ignoring case. Everything else will return `false`. For example, the value `1` will return false. This is inconsistent with the number behavior, where an unexpected value results in a failure.

We should be consistent and throw an exception if the value is not something we expect to get from ServiceNow

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -0,0 +1,138 @@
...
+          return;
+        case BOOLEAN:
+          recordBuilder.set(fieldName, convertToBooleanValue(fieldValue));
+          return;
+        default:
+          throw new IllegalStateException(
+            String.format("Record type '%s' is not supported for field '%s'", fieldType.name(), fieldName));
+      }
+    } catch (ParseException exception) {
+      throw new IllegalStateException(
```
albertshau: IllegalState is normally used for situations that we don't expect to ever happen, like a few lines above when the type is something we don't expect. IllegalArgumentException would be a little better here.

When wrapping exceptions, always include the cause, otherwise the stack trace is lost.

```
throw new IllegalArgumentException(..., exception);
```

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -0,0 +1,138 @@
...
+          try {
+            recordBuilder.setTime(fieldName, LocalTime.parse(fieldValue, timeFormatter));
+          } catch (DateTimeParseException exception) {
+            throw new UnexpectedFormatException(
+              String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
+                            fieldName, fieldSchema.getDisplayName(), fieldValue), exception);
+          }
+          return;
+        default:
+          throw new UnexpectedFormatException(String.format("Field '%s' is of unsupported type '%s'", fieldName,
```
albertshau: change to IllegalStateException to make it consistent with the exception that is thrown when the physical type switch statement

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Model class for API Response from ServiceNow Table APIs
+ */
+public class APIResponse {
+  
+  private List<Map<String, String>> result;
```
albertshau, line 27: How many results will the response have, and how many fields can each result have?

I am concerned about reading the entire response into memory instead of reading it in a streaming fashion.

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public List<Map<String, String>> getResult() {
+    return result;
+  }
+
+  public void setResult(List<Map<String, String>> result) {
```
albertshau: In general, avoid using setters, especially when the class is supposed to represent something immutable. Otherwise it is easy to introduce bugs, where one class reads the value, another class sets it to something else, and now it is unclear what the true state should be.

Instead, variables should all be final and set in a constructor.

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
albertshau: this should be final (see below comment about avoiding setter methods)

src/main/java/io/cdap/plugin/servicenow/sink/model/SchemaResponse.java
```diff
@@ -16,44 +16,21 @@
...
-
-  public void setInternalType(String internalType) {
-    this.internalType = internalType;
-  }
+  
+  private List<ServiceNowSchemaField> result;
 
-  public String getName() {
-    return name;
+  public void setResult(List<ServiceNowSchemaField> result) {
```
albertshau: similar comment to above, set all fields once in a constructor and don't modify state.

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
```diff
@@ -0,0 +1,59 @@
...
+  private String label;
+  private String exampleValue;
+  private String internalType;
+  private String name;
+
+  public String getLabel() {
+    return label;
+  }
+
+  public void setLabel(String label) {
```
albertshau: imilar comment to above, set all fields once in a constructor and don't modify state.

src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
albertshau: TIME_MILLIS will show up in the UI as the physical type (int?) whereas TIME_MICROS will show up as time, so it's better to keep it TIME_MICROS.

albertshau: this also needs to be added to the type mapping doc, I don't see it in the table.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
sgarg-CS: done cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
sgarg-CS: Throwing `UnexpectedFormatException` now for both `convertToDoubleValue()` and `convertToIntegerValue()` methods. cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
sgarg-CS: Throwing `UnexpectedFormatException` for field values other than case-insensitive `Boolean.TRUE` or B`oolean.FALSE`

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
sgarg-CS: Changed to `IllegalArgumentException` and added the cause. cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
sgarg-CS: Changed to `IllegalStateException` cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
sgarg-CS: Added constructor. Removed the setter methods. cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
sgarg-CS: > How many results will the response have, and how many fields can each result have?
> 
> I am concerned about reading the entire response into memory instead of reading it in a streaming fashion.

The ServiceNow Table API response contains a single `result` array. The result array can contain up to 5000 [PAGE_SIZE] records. Each record contains the fields (column names) of the requested table along with their values. The number of fields vary from table to table. Can be as low as ~10 or as high as ~50.

Sample Response Below for sys_user table.
```
{
    "result": [
        {
            "calendar_integration": "1",
            "country": "",
            "last_login_time": "2019-04-05 22:16:30",
            "hashed_user_id": "fdcbd2f17b127236728959a60d57a50338ba8d115adbf14c3d7dffd7dc3ac1fe",
            "source": "",
            "sys_updated_on": "2022-10-04 06:53:43",
            "building": "",
            "web_service_access_only": "false",
            "notification": "2",
            "enable_multifactor_authn": "false",
            "sys_updated_by": "system",
            "sys_created_on": "2019-04-05 21:09:12",
            "sys_domain": {
                "link": "https://ven05127.service-now.com/api/now/table/sys_user_group/global",
                "value": "global"
            },
            "state": "",
            "vip": "false",
            "sys_created_by": "admin",
            "zip": "",
            "home_phone": "",
            "time_format": "",
            "last_login": "2019-04-05",
            "active": "true",
            "sys_domain_path": "/",
            "cost_center": "",
            "phone": "",
            "name": "survey user",
            "employee_number": "",
            "gender": "",
            "city": "",
            "failed_attempts": "0",
            "user_name": "survey.user",
            "title": "",
            "sys_class_name": "sys_user",
            "sys_id": "005d500b536073005e0addeeff7b12f4",
            "internal_integration_user": "false",
            "ldap_server": "",
            "mobile_phone": "",
            "street": "",
            "company": "",
            "department": "",
            "first_name": "survey",
            "email": "survey.user@email.com",
            "introduction": "",
            "preferred_language": "",
            "manager": "",
            "sys_mod_count": "2",
            "last_name": "user",
            "photo": "",
            "avatar": "",
            "middle_name": "",
            "sys_tags": "",
            "time_zone": "",
            "schedule": "",
            "date_format": "",
            "location": ""
        },
        {
            "calendar_integration": "1",
            "country": "US",
            "last_login_time": "",
            "hashed_user_id": "30501c5b4b720ad4fd33232a98f83c21423222e5ce83594e36bf5c657b845431",
            "source": "null",
            "sys_updated_on": "2022-10-06 12:51:33",
            "building": "",
            "web_service_access_only": "false",
            "notification": "2",
            "enable_multifactor_authn": "false",
            "sys_updated_by": "pipeline.user.1",
            "sys_created_on": "2022-10-06 12:51:33",
            "sys_domain": {
                "link": "https://ven05127.service-now.com/api/now/table/sys_user_group/global",
                "value": "global"
            },
            "state": "Massachusetts",
            "vip": "false",
            "sys_created_by": "pipeline.user.1",
            "zip": "2116",
            "home_phone": "",
            "time_format": "",
            "last_login": "",
            "active": "true",
            "sys_domain_path": "/",
            "cost_center": "",
            "phone": "(732) 634-7777",
            "name": "Farmers Insurance ",
            "employee_number": "",
            "gender": "",
            "city": "Boston",
            "failed_attempts": "",
            "user_name": "",
            "title": "",
            "sys_class_name": "sys_user",
            "sys_id": "009e96f287d29110a91edb583cbb3586",
            "internal_integration_user": "false",
            "ldap_server": "",
            "mobile_phone": "",
            "street": "75 Park Plz",
            "company": "",
            "department": "",
            "first_name": "Farmers Insurance",
            "email": "null",
            "introduction": "",
            "preferred_language": "",
            "manager": "",
            "sys_mod_count": "0",
            "last_name": "",
            "photo": "",
            "avatar": "",
            "middle_name": "",
            "sys_tags": "",
            "time_zone": "",
            "schedule": "",
            "date_format": "",
            "location": ""
        }
    ]
}
```


> this should be final (see below comment about avoiding setter methods)
 `result` is now declared as a `final` variable. cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/sink/model/SchemaResponse.java
sgarg-CS: All fields are not being set in the constructor. cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
sgarg-CS: All fields are not being set in the constructor. cbfb212c77242fe61b2ef14c690a9079b6a9e22f

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
albertshau: I don't see page_size being sent anywhere, is that a request property or something else? Also, does that mean the response is actually paginated? I don't see the code going through pages either, it just makes a single request. If the response is paginated, this plugin is definitely not ready for GA, as it will not be able to read anything after the first page.

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
albertshau: remove this setter

src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
albertshau: still have this comment

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
sgarg-CS: The PAGE_SIZE is passed here. https://github.com/cloudsufi/servicenow-plugins/blob/ad60671e737cd75675c454c195419e50d7160720/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java#L102

```
restApi.fetchTableRecordsRetryableMode(tableName, pluginConf.getValueType(), pluginConf.getStartDate(),
                                                     pluginConf.getEndDate(), split.getOffset(),
                                                     ServiceNowConstants.PAGE_SIZE);
```

It later becomes part of the request object 

```
public ServiceNowTableAPIRequestBuilder setLimit(int limit) {
    this.parameters.put("sysparm_limit", String.valueOf(limit));
    return this;
}
```

src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
sgarg-CS: > TIME_MILLIS will show up in the UI as the physical type (int?) whereas TIME_MICROS will show up as time, so it's better to keep it TIME_MICROS.

We found an issue with `TIME_MICROS` during testing where the data did not show up while sampling data through Wrangler. Also had to write additional code for handling `TIME_MICROS`. 
 
 > this also needs to be added to the type mapping doc, I don't see it in the table.

Added the same in the doc. 1e29533aad3428efa26a6a88700f1b9282ae57bd

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
sgarg-CS: Removed. 0ff7d2955ec7a6844c0ee368dec07e59c388c2c8

src/main/java/io/cdap/plugin/servicenow/sink/model/APIResponse.java
albertshau: discussed offline. To close the loop here, the record readers are not paging through the results. Instead, the inputformat creates splits based off the total number of table rows, then sets an offset for each split. So each individual record reader reads one page by design.

pom.xml
```diff
@@ -395,7 +395,7 @@
             <Embed-Directory>lib</Embed-Directory>
             <!--Only @Plugin classes in the export packages will be included as plugin-->
             <_exportcontents>
-              io.cdap.plugin.*;
+              io.cdap.plugin.servicenow.source.*,io.cdap.plugin.servicenow.connector.*;
```
albertshau: I would expect doing this would cause problems where the sink plugin isn't visible. This can instead be `io.cdap.plugin.servicenow.*`

src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
albertshau: discussed this offline, we will keep this time_micros. If there is an issue with it, there will be a separate jira opened with a separate PR.

pom.xml
sgarg-CS: `io.cdap.plugin.servicenow.*`
This makes the ServiceNow Sink Plugin visible in the studio. 

The sink plugin was't ready for release so we skipped it from the export contents.

> I would expect doing this would cause problems where the sink plugin isn't visible. 

Could you please elaborate on some of the problems?

albertshau: problem meaning the sink won't be usable, did not realize that is intentional.

If it is intentional, a better way is to remove the `@Plugin` annotation from the class and add a comment there. That makes it more clear that the plugin is not ready.

albertshau (approved):
lgtm, though please also make the change to remove the plugin annotation from the sink and add a comment about it not being ready.

pom.xml
sgarg-CS: Removed the `@Plugin` annotation and added a comment there about the Sink plugin not being ready yet. 9248f60ddf04f053ff92e673bcf3148d73d2d8ef

sgarg-CS:
Commits have been squashed.

PR #63, opened 2023-10-16
Fix error thrown by connector when 429 response code is received

saimukkamala:
1. When ServiceNow API throws 429 error, response body is HTML (and not json as expected) which causes `JsonSyntaxException` while serializing to Json. This PR addresses this fix and throws a meaningful error.
2. When ServiceNow API errors out when fetching table records, `ServiceNowTableAPIClientImpl.fetchTableRecords` returns  empty list which can cause data loss. Now this will throw an error instead of empty list.

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/RetryableException.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/RetryableException.java
@@ -26,5 +26,9 @@ public class RetryableException extends RuntimeException {
   public RetryableException() {
          super();
   }
+
+    public RetryableException(String message) {
+        super(message);
+    }
   
 }
--- a/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
+++ b/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
@@ -117,6 +117,13 @@ public class RestAPIResponse {
   }
 
   private void checkRetryable() {
+    if (httpStatus == HTTP_STATUS_TOO_MANY_REQUESTS) {
+      isRetryable = true;
+      this.responseBody =
+              String.format(
+                      JSON_ERROR_RESPONSE_TEMPLATE,
+                      "Too many requests to ServiceNow API - decrease concurrent requests");
+    }
     Gson gson = new Gson();
     try {
       JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
@@ -128,19 +135,8 @@ public class RestAPIResponse {
               .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
         isRetryable = true;
       }
-    } catch (JsonSyntaxException e) {
-      // Response Body is not a json object - check status code for error
-      // Response body necessarily need not be json always (when 429 error is thrown -
-      // responseBody is HTML)
-      if (httpStatus == HTTP_STATUS_TOO_MANY_REQUESTS) {
-        isRetryable = true;
-        this.responseBody =
-                String.format(
-                        JSON_ERROR_RESPONSE_TEMPLATE,
-                        "Too many requests to ServiceNow API - decrease concurrent requests");
-      }
     } catch (Throwable t) {
-      // Any other exception
+      // Any exception
       isRetryable = false;
       this.responseBody = String.format(JSON_ERROR_RESPONSE_TEMPLATE, t.getMessage());
     }
--- a/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
+++ b/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
@@ -18,6 +18,7 @@ package io.cdap.plugin.servicenow.restapi;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonSyntaxException;
 import io.cdap.plugin.servicenow.util.ServiceNowConstants;
 import org.apache.http.Header;
 import org.apache.http.HttpResponse;
@@ -48,6 +49,7 @@ public class RestAPIResponse {
     "    },\n" +
     "    \"status\": \"failure\"\n" +
     "}";
+  private static final int HTTP_STATUS_TOO_MANY_REQUESTS = 429;
   private int httpStatus;
   private Map<String, String> headers;
   private String responseBody;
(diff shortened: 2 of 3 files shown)


src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -116,12 +118,31 @@ public boolean isSuccess() {
...
+      if (jo.get(ServiceNowConstants.STATUS) != null
+              && jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE)
+              && jo.getAsJsonObject(ServiceNowConstants.ERROR)
+              .get(ServiceNowConstants.MESSAGE)
+              .getAsString()
+              .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+        isRetryable = true;
+      }
+    } catch (JsonSyntaxException e) {
+      // Response Body is not a json object - check status code for error
```
chtyim: Why don't we check the http status before parsing the body? Typically we should always check the status code before parsing the body.

saimukkamala: Unfortunately servicenow doesn't have any public documentation with an exhaustive list of response codes. For now going with what we know and will be adding them in the future.

src/main/java/io/cdap/plugin/servicenow/apiclient/RetryableException.java
```diff
@@ -26,5 +26,9 @@ public class RetryableException extends RuntimeException {
   public RetryableException() {
          super();
   }
+
+    public RetryableException(String message) {
```
chtyim, line 30: Wrong alignment

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -115,13 +117,28 @@ public boolean isSuccess() {
...
-      jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
-        .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+    if (httpStatus == HTTP_STATUS_TOO_MANY_REQUESTS) {
       isRetryable = true;
+      this.responseBody =
+              String.format(
+                      JSON_ERROR_RESPONSE_TEMPLATE,
+                      "Too many requests to ServiceNow API - decrease concurrent requests");
+    }
+    Gson gson = new Gson();
```
chtyim, line 127: Make it a constant, no need to new it everytime.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -115,13 +117,28 @@ public boolean isSuccess() {
...
   private void checkRetryable() {
-    Gson gson = new Gson();
-    JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
-    if (jo.get(ServiceNowConstants.STATUS) != null &&
-      jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
-      jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
-        .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+    if (httpStatus == HTTP_STATUS_TOO_MANY_REQUESTS) {
       isRetryable = true;
+      this.responseBody =
```
chtyim, line 122: If we know the status is 429, why we need to create a json string, then parse it again just to get the `status` field as `failure`? Seem like we are overly complicating things here.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -48,6 +49,7 @@ public class RestAPIResponse {
     "    },\n" +
     "    \"status\": \"failure\"\n" +
     "}";
+  private static final int HTTP_STATUS_TOO_MANY_REQUESTS = 429;
   private int httpStatus;
```
chtyim, line 53: All fields should be `final`.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -115,13 +117,28 @@ public boolean isSuccess() {
...
-    if (jo.get(ServiceNowConstants.STATUS) != null &&
-      jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
-      jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
-        .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+    if (httpStatus == HTTP_STATUS_TOO_MANY_REQUESTS) {
       isRetryable = true;
+      this.responseBody =
+              String.format(
+                      JSON_ERROR_RESPONSE_TEMPLATE,
+                      "Too many requests to ServiceNow API - decrease concurrent requests");
```
MrRahulSharma, line 125:
```suggestion
                      "ServiceNow Concurrent requests quota reached, please wait for the running requests to complete before triggering any new concurrent request.");
```

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -115,13 +117,28 @@ public boolean isSuccess() {
...
+    }
+    Gson gson = new Gson();
+    try {
+      JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
+      if (jo.get(ServiceNowConstants.STATUS) != null
+              && jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE)
+              && jo.getAsJsonObject(ServiceNowConstants.ERROR)
+              .get(ServiceNowConstants.MESSAGE)
+              .getAsString()
+              .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
```
MrRahulSharma, line 135: Can you check if we require to handle null for the following in the responseBody json ?

1. Missing ERROR node
2. Missing MESSAGE node

PR #64, opened 2023-10-17
ServiceNow Plugin : Make PAGE_SIZE configurable

sgarg-CS:
[PLUGIN-1689](https://cdap.atlassian.net/browse/PLUGIN-1689) : Make Page Size configurable in Service Now Batch Source and Multi-Source Plugins

[PLUGIN-1689]: https://cdap.atlassian.net/browse/PLUGIN-1689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
@@ -69,6 +69,12 @@ public class ServiceNowBaseSourceConfig extends ServiceNowBaseConfig {
     "is set to `Table`.")
   protected String tableNameField;
 
+  @Name(ServiceNowConstants.PROPERTY_PAGE_SIZE)
+  @Macro
+  @Nullable
+  @Description("The number of records to fetch from ServiceNow. Default is 5000.")
+  private Integer pageSize;
+
   /**
    * Constructor for ServiceNowSourceConfig object.
    *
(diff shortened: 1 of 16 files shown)


src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -208,6 +221,17 @@ private void validateDateRange(FailureCollector collector) {
     }
   }
 
+  private void validatePageSize(FailureCollector collector) {
+    if (containsMacro(ServiceNowConstants.PROPERTY_PAGE_SIZE)) {
+      return;
+    }
+
+    if (getPageSize() < 0 || getPageSize() > 10000) {
```
saimukkamala: should be `if (getPageSize() <= 0 || getPageSize() > 10000) {`? `int pages = (tableInfo.getRecordCount() / pageSize);` - this line would divide by 0.

sgarg-CS: Good catch! 
Refactored it. 84fa2da26d0a11988d339f44dd6c112e87cff819

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -208,6 +221,18 @@ private void validateDateRange(FailureCollector collector) {
...
   }
 
+  private void validatePageSize(FailureCollector collector) {
+    if (containsMacro(ServiceNowConstants.PROPERTY_PAGE_SIZE)) {
+      return;
+    }
+
+    if (getPageSize() <= 0 || getPageSize() > 10000) {
+      collector.addFailure("Invalid Page Size.", "Page Size must be greater than 0 and " +
+          "less than 10000.")
```
saimukkamala: less than or equal to 10000

saimukkamala (approved):

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
sgarg-CS: 10000 is a valid Page Size.
Update the failure message. de8166ce721ff42bb0fa51b4d9936fa12f1088e3

PR #66, opened 2023-11-14
Validate Http Response 

saimukkamala:
- Validate Http Response  and throw appropriate errors. 
- Fix silent failing (returns empty results) of Http calls while fetching data.


src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -255,7 +253,7 @@ public Schema fetchTableSchema(String tableName, FailureCollector collector) {
     Schema schema = null;
     try {
       schema = fetchTableSchema(tableName);
-    } catch (OAuthProblemException | OAuthSystemException | RuntimeException e) {
+    } catch (OAuthProblemException | OAuthSystemException | RuntimeException | IOException e) {
```
MrRahulSharma: We can simplify the catch by using catch(Exception), can you check and update ?

sau42shri: @MrRahulSharma In general it is recommended to catch specific exception in lower stack and catch general exception (the ones that are remaining) at caller level, which I guess Sai was doing. However, if you think it is still recommended please go ahead.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -134,15 +134,13 @@ public List<Map<String, String>> fetchTableRecords(String tableName, SourceValue
         if (apiResponse.isRetryable()) {
           throw new RetryableException();
         }
-        return Collections.emptyList();
       }
 
       return parseResponseToResultListOfMap(apiResponse.getResponseBody());
     } catch (OAuthSystemException e) {
       throw new RetryableException();
     } catch (OAuthProblemException e) {
```
sau42shri, line 133: I believe the system vs problem would be around system being busy vs problem being credential being wrong.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -33,31 +36,42 @@
...
+import java.util.Set;
 
 /**
  * An abstract class to call Rest API.
  */
 public abstract class RestAPIClient {
   private static final Logger LOG = LoggerFactory.getLogger(RestAPIClient.class);
+  private static final String HTTP_ERROR_MESSAGE = "Http call to ServiceNow instance returned status code %d";
+  private static final Set<Integer> SUCCESS_CODES = new HashSet<>(Arrays.asList(HttpStatus.SC_OK));
+  private static final Set<Integer> RETRYABLE_CODES = new HashSet<>(Arrays.asList(429,
```
sau42shri: Doesn't HttpStatus have code for 429 as well?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -75,18 +89,15 @@ public RestAPIResponse executePost(RestAPIRequest request) throws IOException {
 
     try (CloseableHttpClient httpClient = HttpClientBuilder.create().build()) {
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpPost)) {
+        validateHttpResponse(httpResponse);
         apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
       }
     } catch (IOException e) {
       // We're retrying all transport exceptions while executing the HTTP POST method and the generic transport
```
sau42shri: This catch block looks like no-op. Do you need it here?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -117,4 +128,15 @@ protected String generateAccessToken(String restApiEndpoint, String clientId, St
     token = client.accessToken(request, OAuth.HttpMethod.POST, OAuthJSONAccessTokenResponse.class).getAccessToken();
     return token;
   }
+
+  private void validateHttpResponse(HttpResponse response) throws IOException {
+    int code = response.getStatusLine().getStatusCode();
+    if (RETRYABLE_CODES.contains(code)) {
+      throw new RetryableException(String.format(HTTP_ERROR_MESSAGE, code));
+    } else if (SUCCESS_CODES.contains(code)) {
```
sau42shri: again no-op piece of code. You could do !SUCCESS_CODES.contains(code) throw new IOException....

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -114,14 +114,18 @@ public boolean isSuccess() {
...
+    try {
+      Gson gson = new Gson();
+      JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
+      if (jo.get(ServiceNowConstants.STATUS) != null &&
+        jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
+        jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
+          .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+        isRetryable = true;
+      }
+    } catch (Exception e) {
```
sau42shri: I am wondering why would this gives a general exception? You are not making an IO call from here. All you are doing is parsing the response body. It should ideally be a parsing error.

sau42shri (commented):
in the process of review... will leave more comments.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
sau42shri: Also as discussed, this should ideally throw some other exception no IOException as it is mainly parsing the response.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -75,18 +89,15 @@ public RestAPIResponse executePost(RestAPIRequest request) throws IOException {
...
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpPost)) {
+        validateHttpResponse(httpResponse);
         apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
       }
     } catch (IOException e) {
       // We're retrying all transport exceptions while executing the HTTP POST method and the generic transport
       // exceptions in HttpClient are represented by the standard java.io.IOException class
       // https://hc.apache.org/httpclient-legacy/exception-handling.html
       throw e;
-    } catch (Exception e) {
```
sau42shri, line 85: Although this is good for current use case, would this be backward incompatible with the existing cdf pipeline as it might throw other exceptions as well?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -118,7 +118,7 @@ private void fetchSchema(ServiceNowTableAPIClientImpl restApi) {
       }
 
       schema = Schema.recordOf(tableName, schemaFields);
-    } catch (OAuthProblemException | OAuthSystemException e) {
```
sau42shri, line 121: Need to consistent like other places. If these are the only ones then fine, but the same should apply for ValidateTable as well.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReaderTest.java
```diff
@@ -198,6 +198,6 @@ public void testFetchDataOnInvalidTable() throws IOException, OAuthProblemExcept
     } catch (RuntimeException | OAuthProblemException | OAuthSystemException e) {
       Assert.assertTrue(e instanceof RuntimeException);
     }
-    Assert.assertFalse(serviceNowMultiRecordReader.nextKeyValue());
```
sau42shri, line 201: I didn't understand this part change. Can you explain?

sau42shri (commented):
added a few more comments.

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
```diff
@@ -146,8 +147,8 @@ public void validateTable(String tableName, SourceValueType valueType, FailureCo
         // Removed config property as in case of MultiSource, only first table error was populating.
         collector.addFailure("Table: " + tableName + " is empty.", "");
       }
-    } catch (OAuthSystemException | OAuthProblemException e) {
-      collector.addFailure("Unable to connect to ServiceNow Instance.",
+    } catch (Exception e) {
+      collector.addFailure(String.format("Unable to connect to ServiceNow Instance. Cause: %s", e.getStackTrace()),
```
chtyim: Don't add stackTrace as a string as it will make the error message very hard to hard and it is not useful for the customer.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -134,15 +134,13 @@ public List<Map<String, String>> fetchTableRecords(String tableName, SourceValue
         if (apiResponse.isRetryable()) {
           throw new RetryableException();
         }
-        return Collections.emptyList();
       }
 
       return parseResponseToResultListOfMap(apiResponse.getResponseBody());
     } catch (OAuthSystemException e) {
       throw new RetryableException();
```
chtyim: Always include the original exception as the cause when rethrowing with a different exception

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -255,7 +253,7 @@ public Schema fetchTableSchema(String tableName, FailureCollector collector) {
     Schema schema = null;
     try {
       schema = fetchTableSchema(tableName);
-    } catch (OAuthProblemException | OAuthSystemException | RuntimeException e) {
+    } catch (Exception e) {
       LOG.error("Error in connection - {}", e.getMessage());
```
chtyim: When logging an exception, always log the whole exception instead of just the message. Also, the message should give more context

```
LOG.error("Failed to fetch schema on table {}", tableName, e);
```

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -33,31 +36,42 @@
...
-  public RestAPIResponse executeGet(RestAPIRequest request) {
+  public RestAPIResponse executeGet(RestAPIRequest request) throws IOException {
     HttpGet httpGet = new HttpGet(request.getUrl());
     request.getHeaders().entrySet().forEach(e -> httpGet.addHeader(e.getKey(), e.getValue()));
     RestAPIResponse apiResponse = null;
 
     try (CloseableHttpClient httpClient = HttpClientBuilder.create().build()) {
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpGet)) {
+        validateHttpResponse(httpResponse);
         apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
```
chtyim: The `apiResponse` local variable is not needed, just do

```
return RestAPIResponse.parse(...);
```

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -75,18 +89,15 @@ public RestAPIResponse executePost(RestAPIRequest request) throws IOException {
 
     try (CloseableHttpClient httpClient = HttpClientBuilder.create().build()) {
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpPost)) {
+        validateHttpResponse(httpResponse);
         apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
```
chtyim: same as above.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
chtyim: It's better to return early to simply the structure.

```
if (SUCCESS_CODES.contains(code)) {
  return;
}
if (RETRYABLE_CODES.contains(code)) {
  throw new RetryableException(....);
}
throw new IOException(...);
```

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -114,14 +114,18 @@ public boolean isSuccess() {
...
-    Gson gson = new Gson();
-    JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
-    if (jo.get(ServiceNowConstants.STATUS) != null &&
-      jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
-      jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
-        .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
-      isRetryable = true;
+  private void checkRetryable() throws IOException {
+    try {
+      Gson gson = new Gson();
```
chtyim: Move this to a static final constant of this class.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -114,14 +114,18 @@ public boolean isSuccess() {
...
+      JsonObject jo = gson.fromJson(this.responseBody, JsonObject.class);
+      if (jo.get(ServiceNowConstants.STATUS) != null &&
+        jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
+        jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
+          .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+        isRetryable = true;
```
chtyim: This method is only called from the constructor. Instead of setting a mutable field, it is better to have this be a static method and return a boolean indicating if it is retryable or not such that the constructor can set it to a final field.

Also, all other fields of this class should be final too.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
saimukkamala: No HttpStatus does not have 429 code.

src/test/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReaderTest.java
saimukkamala: Ideally we should not be calling `recordReader.nextValue()` if `recordReader.initialize()` fails. And it should fail if the the function is called.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
saimukkamala: `OAuthSystemException` is when a system problem occurs, so we retry on this exception. `OAuthProblemException` is when authentication fails.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
saimukkamala: This is not backward incompatible, this exception is caught at a different layer within the connector.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
saimukkamala: That makes sense. Catching exception a layer above.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
saimukkamala: `validateTable` function takes `FailureCollector` as an argument and is expected to not fail and any exception that occurs during function execution should be added to the failure collector.
On the other hand, `RecordReader` functions can fail for any reason and the caller is expected to handle it.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -71,23 +87,20 @@ public RestAPIResponse executePost(RestAPIRequest request) throws IOException {
...
-      // exceptions in HttpClient are represented by the standard java.io.IOException class
-      // https://hc.apache.org/httpclient-legacy/exception-handling.html
-      throw e;
-    } catch (Exception e) {
-      LOG.error("Exception while executing post request", e);
-      apiResponse = RestAPIResponse.defaultErrorResponse(e.getMessage());
     }
-
-    return apiResponse;
+    // We're retrying all transport exceptions while executing the HTTP POST method and the generic transport
```
chtyim: Is this comment block still at the right place? There is no code following this block.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -48,19 +50,19 @@ public class RestAPIResponse {
...
+  private final String responseBody;
+  private final boolean isRetryable;
 
-  public RestAPIResponse(int httpStatus, Map<String, String> headers, String responseBody) {
+  public RestAPIResponse(int httpStatus, Map<String, String> headers, String responseBody) throws IOException {
     this.httpStatus = httpStatus;
     this.headers = headers;
     this.responseBody = responseBody;
-    this.checkRetryable();
+    this.isRetryable = isRetryable(responseBody);
```
chtyim: Shouldn't the `isRetryable` also take `httpStatus` into account?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -36,6 +37,7 @@
  * Pojo class to capture the API response.
  */
 public class RestAPIResponse {
+  private static final Gson gson = new Gson();
```
chtyim: constant should be ALL_CAPS.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -114,15 +109,19 @@ public boolean isSuccess() {
...
+      if (jo.get(ServiceNowConstants.STATUS) != null &&
+        jo.get(ServiceNowConstants.STATUS).getAsString().equals(ServiceNowConstants.FAILURE) &&
+        jo.getAsJsonObject(ServiceNowConstants.ERROR).get(ServiceNowConstants.MESSAGE).getAsString()
+          .contains(ServiceNowConstants.MAXIMUM_EXECUTION_TIME_EXCEEDED)) {
+        return true;
+      }
+    } catch (Exception e) {
+      throw new IOException("Error occurred while reading json response", e);
     }
+    return false;
```
chtyim: You can simplify the logic by just doing the return inside the try block.

```
try {
  JsonObject jo = GSON.fromJson(....);
  return // Your existing if condition;
} catch (Exception e) {
  throw new IOException(..., e);
}
```

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
chtyim: I feel like you should unify the retryable check logic all into this class instead of scattered across this and the RESTApiClient.

saimukkamala: Refactored all functionality related to APIs to `RestAPIResponse` class.
removed `isRetryable` and `isSuccess` variables/methods. `RestAPIResponse.parse()` function returns `RestAPIResponse` object only when there are no errors and throws `RetryableException` or `IOException` depending on the error being retryable or not.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -33,32 +36,39 @@
...
     HttpGet httpGet = new HttpGet(request.getUrl());
     request.getHeaders().entrySet().forEach(e -> httpGet.addHeader(e.getKey(), e.getValue()));
-    RestAPIResponse apiResponse = null;
 
     try (CloseableHttpClient httpClient = HttpClientBuilder.create().build()) {
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpGet)) {
-        apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
+        try {
+          return RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
+        } catch (RetryableException e) {
```
chtyim: This catch and rethrow is unnecessary.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
```diff
@@ -71,23 +81,21 @@ public RestAPIResponse executePost(RestAPIRequest request) throws IOException {
...
 
+    // We're retrying all transport exceptions while executing the HTTP POST method and the generic transport
+    // exceptions in HttpClient are represented by the standard java.io.IOException class
+    // https://hc.apache.org/httpclient-legacy/exception-handling.html
     try (CloseableHttpClient httpClient = HttpClientBuilder.create().build()) {
       try (CloseableHttpResponse httpResponse = httpClient.execute(httpPost)) {
-        apiResponse = RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
+        try {
+          return RestAPIResponse.parse(httpResponse, request.getResponseHeaders());
+        } catch (RetryableException e) {
```
chtyim: Same as above, please remove this catch and rethrow.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIClient.java
chtyim: Now I see the logic better... why you want to wrap all exceptions into IOException besides the retryable one? What exception does the `.parse` method may throw?

MrRahulSharma (approved):

PR #71, opened 2024-03-06
Servicenow Code refactor

MrRahulSharma:


src/e2e-test/java/io/cdap/plugin/servicenowsink/actions/ServiceNowSinkPropertiesPageActions.java
```diff
@@ -50,15 +50,15 @@ public class ServiceNowSinkPropertiesPageActions {
   private static Gson gson = new Gson();
 
   public static void getRecordFromServiceNowTable(String query, String tableName)
-    throws OAuthProblemException, OAuthSystemException {
+    throws OAuthProblemException, OAuthSystemException, IOException {
```
vanathi-g, line 53: Just to clarify, this null represents the pageSize?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -48,23 +49,20 @@ public ServiceNowRecordReader(ServiceNowSourceConfig pluginConf) {
...
   @Override
   public void initialize(InputSplit split, TaskAttemptContext context) {
-    this.split = (ServiceNowInputSplit) split;
-    this.pos = 0;
-    restApi = new ServiceNowTableAPIClientImpl(pluginConf.getConnection());
-    tableName = ((ServiceNowInputSplit) split).getTableName();
-    tableNameField = pluginConf.getTableNameField();
-    fetchSchema(restApi);
+    initialize(split);
+    fetchSchema(new ServiceNowJobConfiguration(context.getConfiguration()).getTableInfos(), tableName);
```
vanathi-g: Little confusing to understand what's going on. fetchSchema takes in a list of tableInfos, gets the schema from the list for given tableName and then initializes the schema field in the RecordReader right?

nit: can get the tableInfos in a separate line and rename the method to fetchAndInitializeSchemas. It would make it a bit more readable.

vanathi-g (approved):
LGTM, minor comments. Also, e2e tests still seem to be failing?

src/e2e-test/java/io/cdap/plugin/servicenowsink/actions/ServiceNowSinkPropertiesPageActions.java
MrRahulSharma: Yes, it got missed in some older change.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
MrRahulSharma: Ack.

MrRahulSharma:
> LGTM, minor comments. Also, e2e tests still seem to be failing?

The failure is due to some data missing from the salesforce setup. Will work on fixing this with Cloudsufi.

PR #88, opened 2025-05-15
[ServiceNow] Support different date, time and datetime formats

anup-cloudsufi:
Support different date, time and datetime formats to handle the ServiceNow datetime parsing issue.

Reference date & time formats: https://www.servicenow.com/docs/bundle/washingtondc-api-reference/page/app-store/dev_portal/API_reference/GlideDateTime/concept/c_GlideDateTimeAPI.html

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
+++ b/src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
@@ -27,6 +27,9 @@ import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
@@ -34,9 +37,50 @@ import java.util.Map;
  * Utility class for converting the record from ServiceNow data type to CDAP schema data types
  */
 public class ServiceNowRecordConverter {
-  private static final String DATE_PATTERN = "yyyy-MM-dd";
-  private static final String DATE_TIME_PATTERN = "yyyy-MM-dd HH:mm:ss";
-  private static final String TIME_PATTERN = "HH:mm:ss";
+
+  // supported date and time formats.
+  // ref: https://www.servicenow.com/docs/bundle/yokohama-api-reference/page/app-store/
+  // dev_portal/API_reference/GlideDateTime/concept/c_GlideDateTimeAPI.html
+  // Instead of using DateTimeFormatterBuilder.appendOptional(...), we maintain a List<DateTimeFormatter>
+  // and attempt to parse the input with each formatter in order. This approach is more reliable because
+  // DateTimeFormatterBuilder processes patterns sequentially and may partially match an incorrect format.
+  // For example, if "MM/dd/yyyy HH:mm:ss" is tried before "dd/MM/yyyy HH:mm:ss", an input like
+  // "14/06/2025 13:00:00" would fail, as 14 is not a valid month.
+  // By explicitly trying each formatter, we avoid such ambiguities and ensure correct parsing of
+  // date formats with overlapping patterns.
+  private static final List<DateTimeFormatter> DATE_TIME_FORMATTER = Collections.unmodifiableList(
+      Arrays.asList(
+        DateTimeFormatter.ofPattern("dd.MM.yyyy hh:mm:ss a"),
+        DateTimeFormatter.ofPattern("dd.MM.yyyy hh.mm.ss a"),
+        DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"),
+        DateTimeFormatter.ofPattern("dd.MM.yyyy HH:mm:ss"),
+        DateTimeFormatter.ofPattern("dd.MM.yyyy HH.mm.ss"),
+        DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss"),
+        DateTimeFormatter.ofPattern("dd-MM-yyyy HH.mm.ss"),
+        DateTimeFormatter.ofPattern("MM/dd/yyyy HH:mm:ss"),
+        DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm:ss"),
+        DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm:ss"),
+        DateTimeFormatter.ofPattern("dd-MM-yy HH.mm.ss"),
+        DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"),
+        DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm"),
+        DateTimeFormatter.ofPattern("dd-MM-yyyy HH.mm")
+      ));
+
+  private static final List<DateTimeFormatter> DATE_FORMATTER = Collections.unmodifiableList(
+    Arrays.asList(
+      DateTimeFormatter.ofPattern("yyyy-MM-dd"),
+      DateTimeFormatter.ofPattern("dd.MM.yyyy"),
+      DateTimeFormatter.ofPattern("dd-MM-yyyy"),
+      DateTimeFormatter.ofPattern("dd/MM/yyyy"),
+      DateTimeFormatter.ofPattern("MM/dd/yyyy"),
+      DateTimeFormatter.ofPattern("MM-dd-yyyy")
+    ));
+
+  private static final List<DateTimeFormatter> TIME_FORMATTER = Collections.unmodifiableList(
+    Arrays.asList(
+      DateTimeFormatter.ofPattern("HH:mm:ss"),
+      DateTimeFormatter.ofPattern("HH:mm")
+    ));
 
   public static void convertToValue(String fieldName, Schema fieldSchema, Map<String, String> record,
                                     StructuredRecord.Builder recordBuilder) {
@@ -53,34 +97,16 @@ public class ServiceNowRecordConverter {
     if (fieldLogicalType != null) {
       switch (fieldLogicalType) {
         case DATETIME:
-          DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern(DATE_TIME_PATTERN);
-          try {
-            recordBuilder.setDateTime(fieldName, LocalDateTime.parse(fieldValue, dateTimeFormatter));
-          } catch (DateTimeParseException exception) {
-            throw new UnexpectedFormatException(
-              String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
-                            fieldName, fieldSchema.getDisplayName(), fieldValue), exception);
-          }
+          recordBuilder.setDateTime(fieldName,
+            parseDateTimeFormat(fieldValue, fieldName, fieldSchema.getDisplayName()));
           return;
         case DATE:
-          DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern(DATE_PATTERN);
-          try {
-            recordBuilder.setDate(fieldName, LocalDate.parse(fieldValue, dateFormatter));
-          } catch (DateTimeParseException exception) {
-            throw new UnexpectedFormatException(
-              String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
-                            fieldName, fieldSchema.getDisplayName(), fieldValue), exception);
-          }
+          recordBuilder.setDate(fieldName,
+            parseDateFormat(fieldValue, fieldName, fieldSchema.getDisplayName()));
           return;
         case TIME_MICROS:
-          DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern(TIME_PATTERN);
-          try {
-            recordBuilder.setTime(fieldName, LocalTime.parse(fieldValue, timeFormatter));
-          } catch (DateTimeParseException exception) {
-            throw new UnexpectedFormatException(
-              String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
-                            fieldName, fieldSchema.getDisplayName(), fieldValue), exception);
-          }
+          recordBuilder.setTime(fieldName,
+            parseTimeFormat(fieldValue, fieldName, fieldSchema.getDisplayName()));
           return;
         default:
           throw new IllegalStateException(String.format("Field '%s' is of unsupported type '%s'", fieldName,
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -20,12 +20,14 @@
 import io.cdap.cdap.api.data.format.UnexpectedFormatException;
 import io.cdap.cdap.api.data.schema.Schema;
 
+
```
itsankit-google: remove empty line

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -34,9 +35,39 @@
...
-  private static final String DATE_PATTERN = "yyyy-MM-dd";
-  private static final String DATE_TIME_PATTERN = "yyyy-MM-dd HH:mm:ss";
-  private static final String TIME_PATTERN = "HH:mm:ss";
+
+  // supported date and time formats.
+  // ref: https://www.servicenow.com/docs/bundle/yokohama-api-reference/page/app-store/
+  // dev_portal/API_reference/GlideDateTime/concept/c_GlideDateTimeAPI.html
+  private static final DateTimeFormatter DATE_TIME_FORMATTER = new DateTimeFormatterBuilder()
+      // Full datetime with seconds
+      .appendOptional(DateTimeFormatter.ofPattern("dd.MM.yyyy hh.mm.ss a"))
```
itsankit-google: referring to the above doc link it also supports `dd.MM.yyyy hh:mm:ss a`, why is this not included?

image

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -34,9 +35,39 @@
...
+      .appendOptional(DateTimeFormatter.ofPattern("dd.MM.yyyy"))
+      .appendOptional(DateTimeFormatter.ofPattern("dd-MM-yyyy"))
+      .appendOptional(DateTimeFormatter.ofPattern("dd/MM/yyyy"))
+      .appendOptional(DateTimeFormatter.ofPattern("MM/dd/yyyy"))
+      .appendOptional(DateTimeFormatter.ofPattern("MM-dd-yyyy"))
+      .toFormatter();
+
+  private static final DateTimeFormatter TIME_FORMATTER = new DateTimeFormatterBuilder()
+      .appendOptional(DateTimeFormatter.ofPattern("HH:mm:ss"))
+      .appendOptional(DateTimeFormatter.ofPattern("HH:mm"))
```
itsankit-google: unable to find this in reference link pasted above, can you please point me to the doc where these are mentioned?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
anup-cloudsufi: added  the missing formats.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
anup-cloudsufi: there is no direct reference to time format in the link, so used the time patterns from the datetime patterns.

itsankit-google: why not `hh:mm:ss a` && `hh.mm.ss a`?

itsankit-google: Also are all of these tested? We should not do guess work in prod plugins.

anup-cloudsufi: Regarding support for `AM/PM` those are majorly used with datetime and not time, hence didn't add it.

Regarding the testing, most of them are tested using the unit test and one of the failure pattern was verified by the customer runs.

itsankit-google: unit testing is one thing but how do we know if they are actually supported by service now if we have added them based on assumption from datetime patterns. Did we do an e2e test with the actual service now data?

prince-cs: removed non-redundant time patterns as confirmed with Servicenow UI, the removed patterns are not required

itsankit-google (changes requested):
I don't see e2e tests for added time formats.

Also I don't see any e2e tests running: 
image

prince-cs:
e2e tests are being enabled for servicenow in https://github.com/data-integrations/servicenow-plugins/pull/89.

itsankit-google:
> e2e tests are being enabled for servicenow in #89.

We need to enable it first, run against this PR. Then only adding any e2e tests make sense.

src/e2e-test/features/servicenowsource/RunTime.feature
```diff
@@ -102,3 +102,54 @@ Feature: ServiceNow Source - Run time scenarios
...
+    And Enter input plugin property: "datasetProject" with value: "datasetprojectId"
+    And Configure BigQuery sink plugin for Dataset and Table
+    Then Validate "BigQuery" plugin properties
+    And Close the Plugin Properties page
+    And Save and Deploy Pipeline
+    And Run the Pipeline in Runtime
+    And Wait till pipeline is in running state
+    And Open and capture logs
+    And Verify the pipeline status is "Succeeded"
+    And Verify count of no of records transferred to the target BigQuery Table
```
itsankit-google: We should also verify actual data other than just records as we do in other plugins e2e tests

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -39,38 +41,47 @@ public class ServiceNowRecordConverter {
...
-      .appendOptional(DateTimeFormatter.ofPattern("HH:mm"))
-      .toFormatter();
+  // We use a List<DateTimeFormatter> and try parsing with each formatter in sequence
+  // instead of using DateTimeFormatterBuilder.appendOptional(...) because:
+  // DateTimeFormatterBuilder.appendOptional(...) processes patterns sequentially,
+  // and if a pattern like "MM/dd/yyyy HH:mm:ss" appears before "dd/MM/yyyy HH:mm:ss",
+  // then an input like "14/06/2025 13:00:00" will incorrectly attempt to parse as MM/dd,
+  // causing a parsing exception since 14 is not a valid month.
+  // By trying each formatter explicitly in code, we avoid such misinterpretations and
+  // can handle ambiguous date formats more reliably.
```
itsankit-google:
```
// Instead of using DateTimeFormatterBuilder.appendOptional(...), we maintain a List<DateTimeFormatter>
// and attempt to parse the input with each formatter in order. This approach is more reliable because
// DateTimeFormatterBuilder processes patterns sequentially and may partially match an incorrect format.
// For example, if "MM/dd/yyyy HH:mm:ss" is tried before "dd/MM/yyyy HH:mm:ss", an input like
// "14/06/2025 13:00:00" would fail, as 14 is not a valid month.
// By explicitly trying each formatter, we avoid such ambiguities and ensure correct parsing of
// date formats with overlapping patterns.
```

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -170,4 +166,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
       String.format("Field with value '%s' is not in valid format.", fieldValue));
     
   }
+
+  @VisibleForTesting
+  private static LocalDateTime parseDateTimeFormat(String fieldValue, String fieldName, String displayName) {
```
itsankit-google: we do not use `@VisibleForTesting` with `private` methods, how is it visible when it is private?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: Done

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: Done

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
       String.format("Field with value '%s' is not in valid format.", fieldValue));
     
   }
+
+  @VisibleForTesting
+  public static LocalDateTime parseDateTimeFormat(String fieldValue, String fieldName, String displayName) {
```
itsankit-google: why is this method public?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
...
+
+      }
+    }
+    throw new UnexpectedFormatException(
+      String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
+        fieldName, displayName, fieldValue));
+  }
+
+  @VisibleForTesting
+  public static LocalDate parseDateFormat(String fieldValue, String fieldName, String displayName) {
```
itsankit-google: why is this method public?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
...
+
+      }
+    }
+    throw new UnexpectedFormatException(
+      String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
+        fieldName, displayName, fieldValue));
+  }
+
+  @VisibleForTesting
+  public static LocalTime parseTimeFormat(String fieldValue, String fieldName, String displayName) {
```
itsankit-google: why is this method public?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
...
+        fieldName, displayName, fieldValue));
+  }
+
+  @VisibleForTesting
+  public static LocalDate parseDateFormat(String fieldValue, String fieldName, String displayName) {
+    for (DateTimeFormatter dateFormatter : DATE_FORMATTER) {
+      try {
+        return LocalDate.parse(fieldValue, dateFormatter);
+      } catch (DateTimeParseException e) {
+
```
itsankit-google: add a comment why this exception is ignored

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
...
+        fieldName, displayName, fieldValue));
+  }
+
+  @VisibleForTesting
+  public static LocalTime parseTimeFormat(String fieldValue, String fieldName, String displayName) {
+    for (DateTimeFormatter timeFormatter : TIME_FORMATTER) {
+      try {
+        return LocalTime.parse(fieldValue, timeFormatter);
+      } catch (DateTimeParseException e) {
+
```
itsankit-google: add a comment why this exception is ignored

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: Other similar methods were marked public so made these methods public as well.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: Other similar methods were marked public so made these methods public as well.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
prince-cs: Other similar methods were marked public so made these methods public as well.

itsankit-google: In Java, methods should be declared private by default unless they need to be accessed from outside the class, to promote encapsulation and minimize unnecessary exposure of implementation details.

prince-cs: Let me make this private then as there is no usage of this method outside of this class.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
...
+      } catch (DateTimeParseException e) {
+        // Only throw exception once all the formats are checked.
+      }
+    }
+    throw new UnexpectedFormatException(
+      String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
+        fieldName, displayName, fieldValue));
+  }
+
+  @VisibleForTesting
```
itsankit-google: Please remove this unnecessary annotation

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
...
```
itsankit-google: Please remove this unnecessary annotation

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowRecordConverter.java
```diff
@@ -139,4 +165,46 @@ public static Boolean convertToBooleanValue(String fieldValue) {
       String.format("Field with value '%s' is not in valid format.", fieldValue));
     
   }
+
+  @VisibleForTesting
```
itsankit-google: Please remove this unnecessary annotation

itsankit-google (changes requested):
Fix e2e test failures.

prince-cs:
e2e tests are being added in https://github.com/data-integrations/servicenow-plugins/pull/92/files

src/e2e-test/features/servicenowsource/RunTime.feature
prince-cs: added in https://github.com/data-integrations/servicenow-plugins/pull/92/files

itsankit-google (approved):

PR #96, opened 2025-06-05
[PLUGIN-1895] Fixing datatype mismatch issue when value type is display

prince-cs:
**Current Issue:**
It has been observed that when the display value type is selected, some values that are defined as integers in the schema appear as strings.

**Code Changes:**
The changes are made on how the schema is fetched. The API associated for fetching the schema is changed to use the column metadata api which now fetches the schema. 
**Schema API:** baseUrl/api/now/doc/table/schema/tableName
**Column Metadata API:** baseUrl/api/now/ui/meta/tableName

In accordance with the response from the column metadata API, the SchemaResponse class is also changed to accommodate the changes. Before mapping the datatype, a check is applied that if the value type is display and if there is a mismatch between the internal_type and type then it needs to be mapped to the type otherwise internal_type.

**Testing:**
The code changes are verified by running unit tests and e2e tests. Also, did run the pipelines for some of the tables for Display and Actual value type.

**JIRA Ticket:** https://cdap.atlassian.net/browse/PLUGIN-1895

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
@@ -48,15 +48,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Predicate;
 import javax.annotation.Nullable;
 
 /**
@@ -247,6 +246,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   }
 
   /**
+   * Fetch schema for actual value type
    * @param tableName ServiceNow table name for which schema is getting fetched
    * @param collector FailureCollector
    * @return schema for given ServiceNow table
@@ -255,7 +255,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   public Schema fetchTableSchema(String tableName, FailureCollector collector) {
     Schema schema = null;
     try {
-      schema = fetchTableSchema(tableName);
+      schema = fetchTableSchema(tableName, SourceValueType.SHOW_ACTUAL_VALUE);
     } catch (Exception e) {
       LOG.error("Failed to fetch schema on table {}", tableName, e);
       collector.addFailure(String.format("Connection failed. Unable to fetch schema for table: %s. Cause: %s",
@@ -276,9 +276,9 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @return schema for given ServiceNow table
    * @throws ServiceNowAPIException
    */
-  public Schema fetchTableSchema(String tableName)
+  public Schema fetchTableSchema(String tableName, SourceValueType valueType)
       throws ServiceNowAPIException {
-      return fetchTableSchema(tableName, getAccessToken());
+      return fetchTableSchema(tableName, getAccessToken(), valueType);
   }
 
   /**
@@ -286,25 +286,32 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    *
    * @param tableName ServiceNow table name for which schema is getting fetched
    * @param accessToken Access Token to use
+   * @param valueType Type of value (Actual/Display)
    * @return schema for given ServiceNow table
    */
-  public Schema fetchTableSchema(String tableName, String accessToken)
+  public Schema fetchTableSchema(String tableName, String accessToken, SourceValueType valueType)
       throws ServiceNowAPIException {
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       this.conf.getRestApiEndpoint(), tableName, true)
       .setExcludeReferenceLink(true);
 
-    RestAPIResponse apiResponse;
+    RestAPIResponse restAPIResponse;
     requestBuilder.setAuthHeader(accessToken);
-    apiResponse = executeGetWithRetries(requestBuilder.build());
-    SchemaResponse response = parseSchemaResponse(apiResponse.getResponseBody());
+    restAPIResponse = executeGetWithRetries(requestBuilder.build());
+    SchemaResponse schemaResponse = parseSchemaResponse(restAPIResponse.getResponseBody());
     List<ServiceNowColumn> columns = new ArrayList<>();
 
-    if (response.getResult() == null && response.getResult().isEmpty()) {
+    if (schemaResponse.getResult() == null && schemaResponse.getResult().getColumns().isEmpty()) {
       throw new RuntimeException("Error - Schema Response does not contain any result");
     }
-    for (ServiceNowSchemaField field : response.getResult()) {
-      columns.add(new ServiceNowColumn(field.getName(), field.getInternalType()));
+
+    for (ServiceNowSchemaField field : schemaResponse.getResult().getColumns().values()) {
+      if (valueType.equals(SourceValueType.SHOW_DISPLAY_VALUE) &&
+        !Objects.equals(field.getType(), field.getInternalType())) {
+        columns.add(new ServiceNowColumn(field.getName(), field.getType()));
+      } else {
+        columns.add(new ServiceNowColumn(field.getName(), field.getInternalType()));
+      }
     }
     return SchemaBuilder.constructSchema(tableName, columns);
   }
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
@@ -43,9 +43,14 @@ public class ServiceNowTableAPIRequestBuilder extends RestAPIRequest.Builder {
    */
   private static final String SCHEMA_API_URL_TEMPLATE = "%s/api/now/doc/table/schema/%s";
 
+  /**
+   * ServiceNow API URL to fetch column metadata
+   */
+  private static final String METADATA_API_URL_TEMPLATE = "%s/api/now/ui/meta/%s";
+
   public ServiceNowTableAPIRequestBuilder(String instanceBaseUrl, String tableName, boolean isSchemaRequired) {
     if (isSchemaRequired) {
-      this.setUrl(String.format(SCHEMA_API_URL_TEMPLATE, instanceBaseUrl, tableName));
+      this.setUrl(String.format(METADATA_API_URL_TEMPLATE, instanceBaseUrl, tableName));
     } else {
       this.setUrl(String.format(TABLE_API_URL_TEMPLATE, instanceBaseUrl, tableName));
     }
--- a/src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
+++ b/src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
@@ -16,7 +16,6 @@
 package io.cdap.plugin.servicenow.connector;
 
 import com.google.gson.Gson;
-import com.google.gson.JsonObject;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
@@ -206,8 +205,14 @@ public class ServiceNowConnector implements DirectConnector {
   @Nullable
   private Schema getSchema(String tableName) {
     SourceQueryMode mode = SourceQueryMode.TABLE;
-    List<ServiceNowTableInfo> tableInfo = ServiceNowInputFormat.fetchTableInfo(mode, config, tableName,
-                                                                               null);
+    // Use display type schema as connector shows a limited number of values
+    // and display value type provides easy to read values
+    List<ServiceNowTableInfo> tableInfo = ServiceNowInputFormat.fetchTableInfo(
+      mode,
+      config,
+      tableName,
+      null,
+      SourceValueType.SHOW_DISPLAY_VALUE);
     Schema schema = tableInfo.stream().findFirst().isPresent() ? tableInfo.stream().findFirst().get().getSchema() :
       null;
     return schema;
--- a/src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
+++ b/src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
@@ -16,28 +16,27 @@
 
 package io.cdap.plugin.servicenow.sink.model;
 
+import com.google.gson.annotations.SerializedName;
+
 /**
  * Model class for Schema Field from Schema API
  */
 public class ServiceNowSchemaField {
   private final String label;
-  private final String exampleValue;
+  @SerializedName("internal_type")
   private final String internalType;
   private final String name;
+  private final String type;
 
-  public ServiceNowSchemaField(String label, String exampleValue, String internalType, String name) {
+  public ServiceNowSchemaField(String label, String internalType, String name, String type) {
     this.label = label;
-    this.exampleValue = exampleValue;
     this.internalType = internalType;
     this.name = name;
+    this.type = type;
   }
 
   public String getLabel() {
     return label;
-  }
-
-  public String getExampleValue() {
-    return exampleValue;
   }
 
   public String getInternalType() {
@@ -46,5 +45,9 @@ public class ServiceNowSchemaField {
 
   public String getName() {
     return name;
+  }
+
+  public String getType() {
+    return type;
   }
 }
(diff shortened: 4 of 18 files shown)


itsankit-google (dismissed):
please add some description in the PR,

with brief explanation under three subheadings:

- What is the current issue?

- What are the changes?

- How the changes are tested?

This will help reviewers understand the change and review PRs efficiently.

Also please create a JIRA and attached to every PR raised to open-source public repos.

itsankit-google:
Also please don;t add reviewers until all these builds are green : 
image

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
```diff
@@ -43,6 +43,8 @@ public class ServiceNowTableAPIRequestBuilder extends RestAPIRequest.Builder {
    */
   private static final String SCHEMA_API_URL_TEMPLATE = "%s/api/now/doc/table/schema/%s";
 
+  private static final String COLUMN_API_URL_TEMPLATE = "%s/api/now/ui/meta/%s";
```
Sunish-Dahiya: I think better name would be METADATA_API_URL_TEMPLATE

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -96,7 +97,7 @@ private static ServiceNowTableInfo getTableMetaData(String tableName, ServiceNow
     Schema schema;
     int recordCount;
     try {
-      schema = restApi.fetchTableSchema(tableName);
+      schema = restApi.fetchTableSchema(tableName, SourceValueType.SHOW_ACTUAL_VALUE);
```
Sunish-Dahiya, line 103: Why are we hardcoding the ValueType, do we not get this value from upstream in this case?

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaResult.java
```diff
@@ -0,0 +1,18 @@
+package io.cdap.plugin.servicenow.sink.model;
+
+import java.util.Map;
+
+/**
+ * Model class for Schema Result from Column Schema API
+ */
+public class ServiceNowSchemaResult {
+  private final Map<String, ServiceNowSchemaField> columns;
```
Sunish-Dahiya, line 54: can you please add more details about the Map, What is its key maybe a example in the comment would be helpful

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
```diff
@@ -16,35 +16,38 @@
...
     this.internalType = internalType;
     this.name = name;
+    this.type = type;
   }
 
   public String getLabel() {
     return label;
   }
 
-  public String getExampleValue() {
```
Sunish-Dahiya, line 39: any reason for removing this field?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
```diff
@@ -43,6 +43,8 @@ public class ServiceNowTableAPIRequestBuilder extends RestAPIRequest.Builder {
...
+
   public ServiceNowTableAPIRequestBuilder(String instanceBaseUrl, String tableName, boolean isSchemaRequired) {
```
Sunish-Dahiya, line 51: if we are not using this constructor anymore why not change the URL here only instead of creating a new constructor?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -48,15 +48,17 @@
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
```
Sunish-Dahiya: Where are we using HashSet and Arrays?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -48,15 +48,17 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
```
Sunish-Dahiya: unused?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -276,35 +278,45 @@ public SchemaResponse parseSchemaResponse(String responseBody) {
...
-    if (response.getResult() == null && response.getResult().isEmpty()) {
+    if (schemaResponse.getResult() == null && schemaResponse.getResult().getColumns().isEmpty()) {
       throw new RuntimeException("Error - Schema Response does not contain any result");
     }
-    for (ServiceNowSchemaField field : response.getResult()) {
-      columns.add(new ServiceNowColumn(field.getName(), field.getInternalType()));
+
+    for (ServiceNowSchemaField field : schemaResponse.getResult().getColumns().values()) {
+      if (valueType.equals(SourceValueType.SHOW_DISPLAY_VALUE)) {
+        if (!Objects.equals(field.getType(), field.getInternalType())) {
```
Sunish-Dahiya: can we combine these two if statements into single?
if(valueType.equals(SourceValueType.SHOW_DISPLAY_VALUE) && !Objects.equals(field.getType(), field.getInternalType())) {
...
}
else {
...
}

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -276,35 +278,45 @@ public SchemaResponse parseSchemaResponse(String responseBody) {
...
     List<ServiceNowColumn> columns = new ArrayList<>();
 
-    if (response.getResult() == null && response.getResult().isEmpty()) {
+    if (schemaResponse.getResult() == null && schemaResponse.getResult().getColumns().isEmpty()) {
       throw new RuntimeException("Error - Schema Response does not contain any result");
     }
-    for (ServiceNowSchemaField field : response.getResult()) {
-      columns.add(new ServiceNowColumn(field.getName(), field.getInternalType()));
+
+    for (ServiceNowSchemaField field : schemaResponse.getResult().getColumns().values()) {
```
Sunish-Dahiya, line 308: This seems like a major change as we are now relying on a new API can we please test and confirm for all the tables that the new schema is exactly same as it was earlier with original API.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
```diff
@@ -176,7 +175,7 @@ private List<StructuredRecord> getTableData(String tableName, int limit)
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       config.getRestApiEndpoint(), tableName, false)
       .setExcludeReferenceLink(true)
-      .setDisplayValue(SourceValueType.SHOW_DISPLAY_VALUE)
+      .setDisplayValue(SourceValueType.SHOW_ACTUAL_VALUE)
```
Sunish-Dahiya: reason for changing the valueType here?

Sunish-Dahiya: Why are we not passing the valueType from config in getTableData instead of hardcoding one value?

Sunish-Dahiya (changes requested):

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
prince-cs: This uses connector config and connector config does not have value type

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaResult.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
prince-cs: This field is not present in the column metadata api being used now to fetch the schema

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
prince-cs: unused imports. Removed now

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
prince-cs: yes. Removed now.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
prince-cs: connector config does not have value type. So, we need to make a call whether to fetch actual or display value type schema.

prince-cs: Displaying actual schema made more sense on the connector level. If you want, I can revert it to display.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
prince-cs: Testing for all the tables is going to be a very time consuming process as there are more than 5800 tables. I can test for some of the major ones. 

Incident and problem table are working fine and generate the correct schema.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
Sunish-Dahiya: What do you mean by connector level? if we change it, will it be backward compatible?

Sunish-Dahiya: Where do we store the valueType then?

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
Sunish-Dahiya: was this field getting used earlier, if yes how are ensuring backward compatibility

Sunish-Dahiya (changes requested):

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
prince-cs: let me revert this code then

prince-cs: `ServiceNowBaseSourceConfig`

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
prince-cs: it is not backward compatible, that's going to be an issue. As the response is changed now, and modal class to accommodate that response is also changed.

Sunish-Dahiya: didn't get it? If this field was getting used earlier then removing it is not the correct solution

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
Sunish-Dahiya: Whats the difference between ServiceNowMultiInputFormat.java and ServiceNowInputFormat.java?
If we are supporting passing valueType in ServiceNowInputFormat why it can't be done in ServiceNowMultiInputFormat?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -255,7 +254,7 @@ public List<Map<String, String>> fetchTableRecordsRetryableMode(String tableName
   public Schema fetchTableSchema(String tableName, FailureCollector collector) {
     Schema schema = null;
     try {
-      schema = fetchTableSchema(tableName);
+      schema = fetchTableSchema(tableName, SourceValueType.SHOW_ACTUAL_VALUE);
```
Sunish-Dahiya, line 258: In the comment can we please mention that this function fetches schema for ACTUAL value type?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: Please attach the test report. i.e. What all tables have been tested, their result and for the change that are not backward compatible how are we ensuring they are working as expected.

Sunish-Dahiya (changes requested):

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
```diff
@@ -206,8 +205,12 @@ private List<StructuredRecord> getTableData(String tableName, int limit)
...
   private Schema getSchema(String tableName) {
     SourceQueryMode mode = SourceQueryMode.TABLE;
-    List<ServiceNowTableInfo> tableInfo = ServiceNowInputFormat.fetchTableInfo(mode, config, tableName,
-                                                                               null);
+    List<ServiceNowTableInfo> tableInfo = ServiceNowInputFormat.fetchTableInfo(
+      mode,
+      config,
+      tableName,
+      null,
+      SourceValueType.SHOW_DISPLAY_VALUE);
```
Sunish-Dahiya, line 215: Please add a comment on why we are hardcoding this value

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
prince-cs: Sorry, my bad. This field had no usage and it is safe to remove it from our codebase.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
prince-cs: already added the test results in the description.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
prince-cs: The function call to fetch the table info is defined in `ServiceNowInputFormat` and that is being used by all the classes. Most of the classes uses connector config object and it does not have value type. So, thats why manual passing of value type is being done.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: I meant in the function description(248), comment here is not adding any value as it is clear from the code itself

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
```diff
@@ -206,8 +205,12 @@ private List<StructuredRecord> getTableData(String tableName, int limit)
...
   private Schema getSchema(String tableName) {
     SourceQueryMode mode = SourceQueryMode.TABLE;
-    List<ServiceNowTableInfo> tableInfo = ServiceNowInputFormat.fetchTableInfo(mode, config, tableName,
-                                                                               null);
+    List<ServiceNowTableInfo> tableInfo = ServiceNowInputFormat.fetchTableInfo(
+      mode,
+      config,
+      tableName,
+      null,
+      SourceValueType.SHOW_DISPLAY_VALUE); // Use display type schema
```
Sunish-Dahiya: Please explain in the comment why are we using SHOW_DISPLAY_VALUE

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
Sunish-Dahiya: Add the comment in code

Sunish-Dahiya (changes requested):

Sunish-Dahiya (approved):

PR #98, opened 2025-06-12
[🍒][PLUGIN-1895] Fixing datatype mismatch issue when value type is display

prince-cs:
cherrypick: https://github.com/data-integrations/servicenow-plugins/pull/96

Current Issue:
It has been observed that when the display value type is selected, some values that are defined as integers in the schema appear as strings.

Code Changes:
The changes are made on how the schema is fetched. The API associated for fetching the schema is changed to use the column metadata api which now fetches the schema.
Schema API: baseUrl/api/now/doc/table/schema/tableName
Column Metadata API: baseUrl/api/now/ui/meta/tableName

In accordance with the response from the column metadata API, the SchemaResponse class is also changed to accommodate the changes. Before mapping the datatype, a check is applied that if the value type is display and if there is a mismatch between the internal_type and type then it needs to be mapped to the type otherwise internal_type.

Testing:
The code changes are verified by running unit tests and e2e tests. Also, did run the pipelines for some of the tables for Display and Actual value type.

JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-1895

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>servicenow-plugins</artifactId>
-  <version>1.2.4-SNAPSHOT</version>
+  <version>1.2.4</version>
   <name>ServiceNow Plugins</name>
   <packaging>jar</packaging>
   <description>Plugins for ServiceNow</description>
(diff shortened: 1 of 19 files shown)


itsankit-google (changes requested):
please add JIRA in the PR title, also add cherrypick in the title.

In the description, please mention the original PR.

For example, see this PR : https://github.com/data-integrations/google-cloud/pull/1553

itsankit-google (approved):

PR #100, opened 2025-06-26
[PLUGIN-1902] Changes done for schema backward compatibility along with metadata API addition for DTS

prince-cs:
As the schema changed logic is changed based on the value type being used. Older pipelines are not compatible with this change. 

Added a flag, which ensures that for DTS users, schema will be fetched via column metadata api and for CDF users it will be fetched via previous schema api.

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
@@ -34,9 +34,12 @@ import io.cdap.plugin.servicenow.restapi.RestAPIClient;
 import io.cdap.plugin.servicenow.restapi.RestAPIResponse;
 import io.cdap.plugin.servicenow.sink.model.APIResponse;
 import io.cdap.plugin.servicenow.sink.model.CreateRecordAPIResponse;
+import io.cdap.plugin.servicenow.sink.model.MetadataAPISchemaResponse;
+import io.cdap.plugin.servicenow.sink.model.SchemaField;
 import io.cdap.plugin.servicenow.sink.model.SchemaResponse;
 import io.cdap.plugin.servicenow.sink.model.ServiceNowSchemaField;
 import io.cdap.plugin.servicenow.util.SchemaBuilder;
+import io.cdap.plugin.servicenow.util.SchemaType;
 import io.cdap.plugin.servicenow.util.ServiceNowColumn;
 import io.cdap.plugin.servicenow.util.ServiceNowConstants;
 import io.cdap.plugin.servicenow.util.SourceValueType;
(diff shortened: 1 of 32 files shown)


src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -289,10 +289,11 @@ public Schema fetchTableSchema(String tableName, SourceValueType valueType)
    * @param valueType Type of value (Actual/Display)
    * @return schema for given ServiceNow table
    */
-  public Schema fetchTableSchema(String tableName, String accessToken, SourceValueType valueType)
+  public Schema fetchTableSchema(String tableName, String accessToken, SourceValueType valueType,
```
vikasrathee-cs, line 313: update the javadoc

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
```diff
@@ -48,9 +48,12 @@ public class ServiceNowTableAPIRequestBuilder extends RestAPIRequest.Builder {
    */
   private static final String METADATA_API_URL_TEMPLATE = "%s/api/now/ui/meta/%s";
 
-  public ServiceNowTableAPIRequestBuilder(String instanceBaseUrl, String tableName, boolean isSchemaRequired) {
-    if (isSchemaRequired) {
+  public ServiceNowTableAPIRequestBuilder(String instanceBaseUrl, String tableName, boolean isSchemaRequired,
+                                          boolean isMetadataAPIRequired) {
+    if (isMetadataAPIRequired) {
       this.setUrl(String.format(METADATA_API_URL_TEMPLATE, instanceBaseUrl, tableName));
+    } else if (isSchemaRequired) {
```
vikasrathee-cs: what is the difference between metadata api vs schema api?

prince-cs: metadata api uses column info which in turn is used to create a schema based on value type and schema api provides with the schema based on actual type only. 

Schema api was being used earlier

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
prince-cs: done

src/main/java/io/cdap/plugin/servicenow/sink/model/SchemaResponseWithoutMetadata.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
```
vikasrathee-cs: change year

vikasrathee-cs:
Reverse the naming convention, let the old schema response and field name class be same and new class can be MetaDataAPISchema and field.

src/main/java/io/cdap/plugin/servicenow/sink/model/SchemaResponseWithoutMetadata.java
prince-cs: done

PR #99, opened 2025-06-20
"[PLUGIN-1901] Time datatype support for actual value type"

Shubhangi-cs:
**Issue:** for fields of type `glide_time` in ServiceNow, when the value is retrieved using the `actual` value type, it is returned in a `glide_datetime` format. This causes a parsing failure, as the expected format for `glide_time` is not maintained. However, when using the `display` value type, the value is correctly returned as `glide_time`.

**Root Cause:** ServiceNow automatically prefixes a default date to `glide_time` values when accessed via the `actual` value type, converting it to `datetime` format.

**Fix:** implemented explicit handling to map `glide_time` fields to `datetime` when using the `actual` value type.

**JIRA Ticket:** https://cdap.atlassian.net/browse/PLUGIN-1901

PR #103, opened 2025-07-08
[cherrypick] [PLUGIN-1901] "Time datatype support for actual value type"

Shubhangi-cs:
CherryPick from https://github.com/data-integrations/servicenow-plugins/pull/99


Sunish-Dahiya (approved):
Please add JIRA to PR title

PR #110, opened 2025-07-17
 [PLUGIN-1902] Changes done for schema backward compatibility along with metadata API addition for DTS

vikasrathee-cs:
[[PLUGIN-1902] Changes done for schema backward compatibility along with metadata API addition for DTS](https://github.com/data-integrations/servicenow-plugins/pull/100)

[PLUGIN-1902]: https://cdap.atlassian.net/browse/PLUGIN-1902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
+++ b/src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
@@ -28,13 +28,10 @@ import io.cdap.plugin.servicenow.apiclient.ServiceNowTableAPIRequestBuilder;
 import io.cdap.plugin.servicenow.connector.ServiceNowConnectorConfig;
 import io.cdap.plugin.servicenow.restapi.RestAPIResponse;
 import io.cdap.plugin.servicenow.source.ServiceNowSourceConfig;
+import io.cdap.plugin.servicenow.util.SchemaType;
 import io.cdap.plugin.servicenow.util.ServiceNowConstants;
 import io.cdap.plugin.servicenow.util.SourceValueType;
-import org.apache.http.HttpStatus;
-import org.apache.oltu.oauth2.common.exception.OAuthProblemException;
-import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
 
-import java.io.IOException;
 import javax.annotation.Nullable;
 
 /**
@@ -56,6 +53,11 @@ public class ServiceNowBaseConfig extends PluginConfig {
   public ServiceNowBaseConfig(String clientId, String clientSecret, String restApiEndpoint,
                               String user, String password) {
     this.connection = new ServiceNowConnectorConfig(clientId, clientSecret, restApiEndpoint, user, password);
+  }
+
+  @Nullable
+  public Boolean getUseConnection() {
+    return useConnection;
   }
 
   @Nullable
@@ -83,7 +85,7 @@ public class ServiceNowBaseConfig extends PluginConfig {
   @VisibleForTesting
   public void validateServiceNowConnection(FailureCollector collector) {
     try {
-      ServiceNowTableAPIClientImpl restApi = new ServiceNowTableAPIClientImpl(connection);
+      ServiceNowTableAPIClientImpl restApi = new ServiceNowTableAPIClientImpl(connection, useConnection);
       restApi.getAccessToken();
     } catch (Exception e) {
       collector.addFailure("Unable to connect to ServiceNow Instance.",
@@ -123,13 +125,13 @@ public class ServiceNowBaseConfig extends PluginConfig {
                             String tableField) {
     // Call API to fetch first record from the table
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
-      connection.getRestApiEndpoint(), tableName, false)
+      connection.getRestApiEndpoint(), tableName, false, SchemaType.SCHEMA_API_BASED)
       .setExcludeReferenceLink(true)
       .setDisplayValue(valueType)
       .setLimit(1);
 
     RestAPIResponse apiResponse = null;
-    ServiceNowTableAPIClientImpl serviceNowTableAPIClient = new ServiceNowTableAPIClientImpl(connection);
+    ServiceNowTableAPIClientImpl serviceNowTableAPIClient = new ServiceNowTableAPIClientImpl(connection, useConnection);
     try {
       String accessToken = serviceNowTableAPIClient.getAccessToken();
       requestBuilder.setAuthHeader(accessToken);
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
@@ -30,13 +30,16 @@ import com.google.gson.reflect.TypeToken;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.servicenow.connector.ServiceNowConnectorConfig;
+import io.cdap.plugin.servicenow.model.APIResponse;
+import io.cdap.plugin.servicenow.model.CreateRecordAPIResponse;
+import io.cdap.plugin.servicenow.model.MetadataAPISchemaField;
+import io.cdap.plugin.servicenow.model.MetadataAPISchemaResponse;
+import io.cdap.plugin.servicenow.model.SchemaAPISchemaField;
+import io.cdap.plugin.servicenow.model.SchemaAPISchemaResponse;
 import io.cdap.plugin.servicenow.restapi.RestAPIClient;
 import io.cdap.plugin.servicenow.restapi.RestAPIResponse;
-import io.cdap.plugin.servicenow.sink.model.APIResponse;
-import io.cdap.plugin.servicenow.sink.model.CreateRecordAPIResponse;
-import io.cdap.plugin.servicenow.sink.model.SchemaResponse;
-import io.cdap.plugin.servicenow.sink.model.ServiceNowSchemaField;
 import io.cdap.plugin.servicenow.util.SchemaBuilder;
+import io.cdap.plugin.servicenow.util.SchemaType;
 import io.cdap.plugin.servicenow.util.ServiceNowColumn;
 import io.cdap.plugin.servicenow.util.ServiceNowConstants;
 import io.cdap.plugin.servicenow.util.SourceValueType;
@@ -76,10 +79,12 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   private static final String GLIDE_DATE_TIME_DATATYPE = "glide_date_time";
   private static final Gson GSON = new Gson();
   private final ServiceNowConnectorConfig conf;
+  public final SchemaType schemaType;
   public static JsonArray serviceNowJsonResultArray;
 
-  public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf) {
+  public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf, Boolean useConnection) {
     this.conf = conf;
+    this.schemaType = getSchemaTypeBasedOnUseConnection(useConnection);
   }
 
   public String getAccessToken() throws ServiceNowAPIException {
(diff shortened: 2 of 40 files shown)


src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -76,10 +79,12 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   private static final String GLIDE_DATE_TIME_DATATYPE = "glide_date_time";
   private static final Gson GSON = new Gson();
   private final ServiceNowConnectorConfig conf;
+  public final Boolean useConnection;
   public static JsonArray serviceNowJsonResultArray;
 
-  public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf) {
+  public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf, Boolean useConnection) {
     this.conf = conf;
+    this.useConnection = useConnection;
```
sahusanket: Since this class is an implementation, is should be tied to a Single SchemaType.

I do not see the use of  `useConnection` for any thing else. 

So, instead of storing and recomputing again n again, Can we just store `SchemaType` as final field?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -271,20 +275,37 @@ public Schema fetchTableSchema(String tableName, FailureCollector collector) {
...
    *
    * @param tableName ServiceNow table name for which schema is getting fetched
+   * @param valueType Whether to fetch schema for actual value or display value
+   * @param useConnection This flag is added to backward compatibility with 1.1 release,
+   *                      in case of null for useConnection, schema will be string based.
    * @return schema for given ServiceNow table
    * @throws ServiceNowAPIException
    */
-  public Schema fetchTableSchema(String tableName, SourceValueType valueType)
+  public Schema fetchTableSchema(String tableName, SourceValueType valueType, Boolean useConnection)
```
sahusanket: Do you see a usecase where we need to get `useConnection` value that would be different from that we are taking in constructor ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -271,20 +275,37 @@ public Schema fetchTableSchema(String tableName, FailureCollector collector) {
   }
 
   @VisibleForTesting
-  public SchemaResponse parseSchemaResponse(String responseBody) {
+  public MetadataAPISchemaResponse parseSchemaResponse(String responseBody) {
```
sahusanket, line 278: I see this is called from a private method Once . 
Can we make it private  or just move it there, since it's just 1 line ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -271,20 +275,37 @@ public Schema fetchTableSchema(String tableName, FailureCollector collector) {
...
+    return GSON.fromJson(responseBody, MetadataAPISchemaResponse.class);
+  }
+
+  public SchemaResponse parseSchemaResponseWithoutMetadata(String responseBody) {
```
sahusanket: same : 
I see this is called from a private method Once . 
Can we make it private  or just move it there, since it's just 1 line ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -293,25 +314,88 @@ public Schema fetchTableSchema(String tableName, SourceValueType valueType)
...
+   * @return A {@link Schema} object representing the table structure as interpreted from the Schema API.
+   *
+   * @throws RuntimeException if the schema response is null or contains no result.
+   */
+  private Schema prepareSchemaWithSchemaAPI(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
+                                            String tableName) {
+    SchemaResponse schemaResponse =
+      parseSchemaResponseWithoutMetadata(restAPIResponse.getResponseBody());
+
+    if (schemaResponse.getResult() == null && schemaResponse.getResult().isEmpty()) {
```
sahusanket: This will throw NPE.. if null 

it should be or `||`

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -293,25 +314,88 @@ public Schema fetchTableSchema(String tableName, SourceValueType valueType)
...
+   *
+   * @return A {@link Schema} object representing the table structure as interpreted from the Metadata API.
+   *
+   * @throws RuntimeException if the response does not contain valid column information.
+   */
+  private Schema prepareSchemaWithMetadataAPI(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
+                                              String tableName, SourceValueType valueType) {
+    MetadataAPISchemaResponse metadataAPISchemaResponse = parseSchemaResponse(restAPIResponse.getResponseBody());
+
+    if (metadataAPISchemaResponse.getResult() == null && metadataAPISchemaResponse.getResult().getColumns().isEmpty()) {
```
sahusanket: same 
This will throw NPE.. if null

it should be or `||`

sahusanket:
Any reason why `model` is inside sink ? Is it not used by validation and source? 

If no strong reason then please refactor it to `io.cdap.plugin.servicenow.model`

src/main/java/io/cdap/plugin/servicenow/sink/model/ServiceNowSchemaField.java
```diff
@@ -19,7 +19,7 @@
 import com.google.gson.annotations.SerializedName;
 
 /**
- * Model class for Schema Field from Schema API
+ * Model class for Schema Field from Column Metadata API
  */
 public class ServiceNowSchemaField {
```
sahusanket: There is SchemaField and Service now schema field. 

Please Rename both to indicate what they are dedicated to. 

Example 
MetadataApiSchemaField 
SchemaApiSchemaField ( or anything better )

sahusanket:
I see a lot of Old code that was used for SCHEMA API , is not used to MetaData APi instead ? 

Is that expected? and why ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
vikasrathee-cs: It is getting called from unit test cases also

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
vikasrathee-cs: removed

vikasrathee-cs:
> I see a lot of Old code that was used for SCHEMA API , is not used to MetaData API instead ?
> Is that expected? and why ?

Previously Schema API was removed and same classes were used for Metadata API, later due to backward compatibility we had to add schema api again. Renamed the classes again to provide the clarity on which class getting used by which api.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -293,25 +309,92 @@ public Schema fetchTableSchema(String tableName, SourceValueType valueType)
...
 
     RestAPIResponse restAPIResponse;
     requestBuilder.setAuthHeader(accessToken);
     restAPIResponse = executeGetWithRetries(requestBuilder.build());
-    SchemaResponse schemaResponse = parseSchemaResponse(restAPIResponse.getResponseBody());
     List<ServiceNowColumn> columns = new ArrayList<>();
 
-    if (schemaResponse.getResult() == null && schemaResponse.getResult().getColumns().isEmpty()) {
-      throw new RuntimeException("Error - Schema Response does not contain any result");
+    if (schemaType == SchemaType.METADATA_API_BASED) {
```
sahusanket, line 325: In case of DTS it directly calls this method ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIRequestBuilder.java
```diff
@@ -48,9 +49,17 @@ public class ServiceNowTableAPIRequestBuilder extends RestAPIRequest.Builder {
    */
   private static final String METADATA_API_URL_TEMPLATE = "%s/api/now/ui/meta/%s";
 
-  public ServiceNowTableAPIRequestBuilder(String instanceBaseUrl, String tableName, boolean isSchemaRequired) {
+  public ServiceNowTableAPIRequestBuilder(String instanceBaseUrl, String tableName, boolean isSchemaRequired,
+                                          SchemaType schemaType) {
     if (isSchemaRequired) {
-      this.setUrl(String.format(METADATA_API_URL_TEMPLATE, instanceBaseUrl, tableName));
+      if (SchemaType.METADATA_API_BASED == schemaType) {
```
sahusanket, line 59: Please simplify this 

move `this.setUrl(String.format(TABLE_API_URL_TEMPLATE, instanceBaseUrl, tableName));` before the `if` block  as the default

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -76,10 +79,14 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
...
   private static final Gson GSON = new Gson();
   private final ServiceNowConnectorConfig conf;
+  public final Boolean useConnection;
+  public final SchemaType schemaType;
   public static JsonArray serviceNowJsonResultArray;
 
-  public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf) {
+  public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf, Boolean useConnection) {
     this.conf = conf;
+    this.useConnection = useConnection;
```
sahusanket: Do we still need this field ?

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
```diff
@@ -123,13 +125,13 @@ public void validateTable(String tableName, SourceValueType valueType, FailureCo
                             String tableField) {
     // Call API to fetch first record from the table
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
-      connection.getRestApiEndpoint(), tableName, false)
+      connection.getRestApiEndpoint(), tableName, false, SchemaType.SCHEMA_API_BASED)
```
sahusanket, line 128: Why is this always SCHEMA_API ? 

What happens in case of 1.1 ( string based ) ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
vikasrathee-cs: yes

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
vikasrathee-cs: no, removed.

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
vikasrathee-cs: this is not a schema call, as previous parameter is showing false. Thats why we are setting default schema type

sahusanket (approved):

sahusanket:
Please squash before merging.

PR #131, opened 2025-11-10
PLUGIN-1936: Read data in a streaming manner

sgarg-CS:
[PLUGIN-1936](https://cdap.atlassian.net/browse/PLUGIN-1936)

**Issue** : OutOfMemory (OOM) Errors seen in the ServiceNow Connector frequently when transfering large number of records

**RCA**: The entire response is converted into string and the JSON is parsed in memory leading to OOM issues. 

**Proposed Fix:**  Read the records incrementally i.e. one by one in a streaming manner. 

**Changes** : 
- Refactored the existing method: `prepareResponse` :in the `RESTAPIResponse` class to return a reusable InputStream in the `RESTAPIResponse` object. 
- Added `openNextPage` to initialise the `Gson's` `JsonReader` and read the `result` array in the response and `closeCurrentPage` to close the `JsonReader`
- `nextKeyValue()` : Refactored the existing logic to stream one record at a time.
- Changed the return type of the following methods from Map<String, String> to RESTAPIResponse
   - `fetchData()`
   - `fetchTableRecordsRetryableMode()`
   - `fetchTableRecords()`
   
**Testing**
image

  

[PLUGIN-1936]: https://cdap.atlassian.net/browse/PLUGIN-1936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
+++ b/src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
@@ -17,6 +17,7 @@
 package io.cdap.plugin.servicenow;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.stream.JsonReader;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
@@ -31,7 +32,11 @@ import io.cdap.plugin.servicenow.source.ServiceNowSourceConfig;
 import io.cdap.plugin.servicenow.util.SchemaType;
 import io.cdap.plugin.servicenow.util.ServiceNowConstants;
 import io.cdap.plugin.servicenow.util.SourceValueType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.io.InputStreamReader;
 import javax.annotation.Nullable;
 
 /**
@@ -39,6 +44,7 @@ import javax.annotation.Nullable;
  */
 public class ServiceNowBaseConfig extends PluginConfig {
 
+  private static final Logger log = LoggerFactory.getLogger(ServiceNowBaseConfig.class);
   @Name(ConfigUtil.NAME_USE_CONNECTION)
   @Nullable
   @Description("Whether to use an existing connection.")
@@ -140,7 +146,7 @@ public class ServiceNowBaseConfig extends PluginConfig {
       requestBuilder.setResponseHeaders(ServiceNowConstants.HEADER_NAME_TOTAL_COUNT);
 
       apiResponse = serviceNowTableAPIClient.executeGetWithRetries(requestBuilder.build());
-      if (serviceNowTableAPIClient.parseResponseToResultListOfMap(apiResponse.getResponseBody()).isEmpty()) {
+      if (isResultEmpty(apiResponse)) {
         // Removed config property as in case of MultiSource, only first table error was populating.
         collector.addFailure("Table: " + tableName + " is empty.", "");
       }
@@ -150,6 +156,33 @@ public class ServiceNowBaseConfig extends PluginConfig {
                            "Ensure specified table exists in the datasource. ")
         .withConfigProperty(ServiceNowConstants.PROPERTY_TABLE_NAME);
     }
+  }
+
+  /**
+   * Checks if the "result" array in the ServiceNow REST API response is empty.
+   * <p>
+   * Determines if the "result" array in a ServiceNow REST API response is empty by specifically looking for a top-level
+   * key named "result". Once found, it opens the associated array and checks for the presence of a first element.
+   * </p>
+   * 
+   * @param restAPIResponse The response object containing the JSON input stream
+   * @return true, if the "result" array exists and is empty, or if the "result" key is never found;
+   * false, if the array contains at least one element.
+   * @throws IOException If there is an error reading the input stream or parsing the JSON.
+   */
+  public boolean isResultEmpty(RestAPIResponse restAPIResponse) throws IOException {
+    JsonReader reader = new JsonReader(new InputStreamReader(restAPIResponse.getResponseStream()));
+    reader.beginObject();
+    while (reader.hasNext()) {
+      String name = reader.nextName();
+      if (ServiceNowConstants.RESULT.equals(name)) {
+        reader.beginArray();
+        return !reader.hasNext();
+      } else {
+        reader.skipValue();
+      }
+    }
+    return true;
   }
 
 }
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
@@ -27,7 +27,7 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
-import com.sun.org.apache.xpath.internal.operations.Bool;
+import com.google.gson.stream.JsonReader;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.servicenow.connector.ServiceNowConnectorConfig;
@@ -56,7 +56,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -86,6 +89,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   public static JsonArray serviceNowJsonResultArray;
 
   public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf, Boolean useConnection) {
+    super();
     this.conf = conf;
     this.schemaType = getSchemaTypeBasedOnUseConnection(useConnection);
   }
@@ -135,7 +139,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @return A list of maps, where each map represents a record from the table.
    * @throws ServiceNowAPIException If an error occurs while fetching the records.
    */
-  public List<Map<String, String>> fetchTableRecords(
+  public RestAPIResponse fetchTableRecords(
     String tableName,
     SourceValueType valueType,
     String filterQuery,
@@ -148,6 +152,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
       .setDisplayValue(valueType)
       .setLimit(limit);
 
+    requestBuilder.setResponseHeaders(ServiceNowConstants.HEADER_NAME_TOTAL_COUNT);
+
     if (offset > 0) {
       requestBuilder.setOffset(offset);
     }
@@ -159,7 +165,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
     String accessToken = getAccessToken();
     requestBuilder.setAuthHeader(accessToken);
     RestAPIResponse apiResponse = executeGetWithRetries(requestBuilder.build());
-    return parseResponseToResultListOfMap(apiResponse.getResponseBody());
+    return apiResponse;
   }
 
   private int getRecordCountFromHeader(RestAPIResponse apiResponse) {
@@ -209,30 +215,24 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @param limit     The number of records to be fetched
    * @return The list of Map; each Map representing a table row
    */
-  public List<Map<String, String>> fetchTableRecordsRetryableMode(String tableName, SourceValueType valueType,
+  public RestAPIResponse fetchTableRecordsRetryableMode(String tableName, SourceValueType valueType,
                                                                   String filterQuery, int offset,
                                                                   int limit) throws ServiceNowAPIException {
-    final List<Map<String, String>> results = new ArrayList<>();
-    Callable<Boolean> fetchRecords = () -> {
-      results.addAll(fetchTableRecords(tableName, valueType, filterQuery, offset, limit));
-      return true;
-    };
+    Callable<RestAPIResponse> fetchRecords = () -> fetchTableRecords(tableName, valueType, filterQuery, offset, limit);
 
-    Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
+    Retryer<RestAPIResponse> retryer = RetryerBuilder.<RestAPIResponse>newBuilder()
       .retryIfException(this::isExceptionRetryable)
       .withWaitStrategy(WaitStrategies.exponentialWait(ServiceNowConstants.WAIT_TIME, TimeUnit.MILLISECONDS))
       .withStopStrategy(StopStrategies.stopAfterAttempt(ServiceNowConstants.MAX_NUMBER_OF_RETRY_ATTEMPTS))
       .build();
 
     try {
-      retryer.call(fetchRecords);
+      return retryer.call(fetchRecords);
     } catch (RetryException | ExecutionException e) {
       throw new ServiceNowAPIException(
           String.format("Data Recovery failed for batch %s to %s.", offset, (offset + limit)),
           e, null, false);
     }
-
-    return results;
   }
 
   /**
@@ -255,8 +255,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   }
 
   @VisibleForTesting
-  public MetadataAPISchemaResponse parseSchemaResponse(String responseBody) {
-    return GSON.fromJson(responseBody, MetadataAPISchemaResponse.class);
+  public MetadataAPISchemaResponse parseSchemaResponse(InputStream responseStream) {
+    return GSON.fromJson(createJsonReader(responseStream), MetadataAPISchemaResponse.class);
   }
 
   /**
@@ -268,8 +268,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @throws ServiceNowAPIException
    */
   public Schema fetchTableSchema(String tableName, SourceValueType valueType)
-      throws ServiceNowAPIException {
-    return fetchTableSchema(tableName, getAccessToken(), valueType, schemaType, false);
+    throws ServiceNowAPIException {
+    return fetchTableSchema(tableName, getAccessToken(), valueType, schemaType, true);
   }
 
   private SchemaType getSchemaTypeBasedOnUseConnection(Boolean useConnection) {
@@ -293,7 +293,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    */
   public Schema fetchTableSchema(String tableName, String accessToken, SourceValueType valueType,
                                  SchemaType schemaType, Boolean legacyMapping)
-      throws ServiceNowAPIException {
+    throws ServiceNowAPIException {
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       this.conf.getRestApiEndpoint(), tableName, true, schemaType)
       .setExcludeReferenceLink(true);
@@ -303,12 +303,16 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
     restAPIResponse = executeGetWithRetries(requestBuilder.build());
     List<ServiceNowColumn> columns = new ArrayList<>();
 
-    if (schemaType == SchemaType.METADATA_API_BASED) {
-      return prepareSchemaWithMetadataAPI(restAPIResponse, columns, tableName, valueType, legacyMapping);
-    } else if (schemaType == SchemaType.SCHEMA_API_BASED) {
-      return prepareSchemaWithSchemaAPI(restAPIResponse, columns, tableName);
-    } else {
-      return prepareStringBasedSchema(restAPIResponse, columns, tableName);
+    try {
+      if (schemaType == SchemaType.METADATA_API_BASED) {
+        return prepareSchemaWithMetadataAPI(restAPIResponse, columns, tableName, valueType, legacyMapping);
+      } else if (schemaType == SchemaType.SCHEMA_API_BASED) {
+        return prepareSchemaWithSchemaAPI(restAPIResponse, columns, tableName, legacyMapping);
+      } else {
+        return prepareStringBasedSchema(restAPIResponse, columns, tableName, legacyMapping);
+      }
+    } catch (IOException exception) {
+      throw new RuntimeException("Error in fetching schema for table " + tableName, exception);
     }
   }
 
@@ -329,9 +333,9 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @throws RuntimeException if the schema response is null or contains no result.
    */
   private Schema prepareSchemaWithSchemaAPI(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
-                                            String tableName) throws ServiceNowAPIException {
+    String tableName, Boolean legacyMapping) throws ServiceNowAPIException, IOException {
     SchemaAPISchemaResponse schemaAPISchemaResponse =
-      GSON.fromJson(restAPIResponse.getResponseBody(), SchemaAPISchemaResponse.class);
+      GSON.fromJson(createJsonReader(restAPIResponse.getResponseStream()), SchemaAPISchemaResponse.class);
 
     if (schemaAPISchemaResponse.getResult() == null || schemaAPISchemaResponse.getResult().isEmpty()) {
       throw new ServiceNowAPIException(
@@ -341,7 +345,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
     for (SchemaAPISchemaField field : schemaAPISchemaResponse.getResult()) {
       columns.add(new ServiceNowColumn(field.getName(), field.getInternalType()));
     }
-    return SchemaBuilder.constructSchema(tableName, columns, false);
+    return SchemaBuilder.constructSchema(tableName, columns, legacyMapping);
   }
 
   /**
@@ -363,8 +367,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @throws RuntimeException if the response does not contain valid column information.
    */
   private Schema prepareSchemaWithMetadataAPI(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
-    String tableName, SourceValueType valueType, Boolean legacyMapping) throws ServiceNowAPIException {
-    MetadataAPISchemaResponse metadataAPISchemaResponse = parseSchemaResponse(restAPIResponse.getResponseBody());
+    String tableName, SourceValueType valueType, Boolean legacyMapping) throws ServiceNowAPIException, IOException {
+    MetadataAPISchemaResponse metadataAPISchemaResponse = parseSchemaResponse(restAPIResponse.getResponseStream());
 
     if (metadataAPISchemaResponse.getResult() == null || metadataAPISchemaResponse.getResult().getColumns() == null ||
       metadataAPISchemaResponse.getResult().getColumns().isEmpty()) {
@@ -389,6 +393,11 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
       }
     }
     return SchemaBuilder.constructSchema(tableName, columns, legacyMapping);
+  }
+
+  public JsonReader createJsonReader(InputStream inputStream) {
+    Objects.requireNonNull(inputStream, "InputStream must not be null");
+    return new JsonReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
   }
 
   /**
@@ -451,8 +460,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
       requestBuilder.setContentTypeHeader("application/json");
       requestBuilder.setEntity(entity);
       apiResponse = executePost(requestBuilder.build());
-
       systemID = String.valueOf(getSystemId(apiResponse));
+      apiResponse.close();
     } catch (IOException e) {
       throw new ServiceNowAPIException("Error in creating a new record", e, null, false);
     }
@@ -482,15 +491,17 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
       apiResponse = executePost(requestBuilder.build());
 
       systemID = String.valueOf(getSystemId(apiResponse));
+      apiResponse.close();
     } catch (IOException e) {
       throw new ServiceNowAPIException("Error in creating a new record", e, null, false);
     }
     return systemID;
   }
 
-  private String getSystemId(RestAPIResponse restAPIResponse) {
-    CreateRecordAPIResponse apiResponse = GSON.fromJson(restAPIResponse.getResponseBody(),
-                                                           CreateRecordAPIResponse.class);
+  private String getSystemId(RestAPIResponse restAPIResponse) throws IOException {
+    CreateRecordAPIResponse apiResponse = GSON.fromJson(
+      new InputStreamReader(restAPIResponse.getResponseStream(), StandardCharsets.UTF_8),
+        CreateRecordAPIResponse.class);
     return apiResponse.getResult().get(ServiceNowConstants.SYSTEM_ID).toString();
   }
 
@@ -501,8 +512,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    * @param tableName The ServiceNow table name
    * @param query The query
    */
-  public Map<String, String> getRecordFromServiceNowTable(String tableName, String query)
-      throws ServiceNowAPIException {
+  public JsonObject getRecordFromServiceNowTable(String tableName, String query)
+    throws ServiceNowAPIException, IOException {
 
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       this.conf.getRestApiEndpoint(), tableName, false, schemaType)
@@ -513,7 +524,8 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
     requestBuilder.setAuthHeader(accessToken);
     restAPIResponse = executeGetWithRetries(requestBuilder.build());
 
-    APIResponse apiResponse = GSON.fromJson(restAPIResponse.getResponseBody(), APIResponse.class);
+    APIResponse apiResponse = GSON.fromJson(
+      new InputStreamReader(restAPIResponse.getResponseStream(), StandardCharsets.UTF_8), APIResponse.class);
     return apiResponse.getResult().get(0);
   }
 
@@ -530,16 +542,42 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
    *
    * @throws RuntimeException if the schema response is null or contains no result.
    */
-  private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
-                                          String tableName) {
-    List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getResponseBody());
-    if (result != null && !result.isEmpty()) {
-      Map<String, String> firstRecord = result.get(0);
-      for (String key : firstRecord.keySet()) {
-        columns.add(new ServiceNowColumn(key, "string"));
+    private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
+                                          String tableName, Boolean legacyMapping) throws IOException {
+      InputStream in = restAPIResponse.getResponseStream();
+      JsonReader reader = new JsonReader(new InputStreamReader(in, StandardCharsets.UTF_8));
+      JsonObject firstRecord;
+      try {
+        firstRecord = getFirstRecord(reader);
+        restAPIResponse.close();
+      } catch (IOException e) {
+        throw new RuntimeException("Failed to parse schema for table: " + tableName, e);
       }
-      return SchemaBuilder.constructSchema(tableName, columns, false);
+      if (firstRecord != null) {
+      firstRecord.entrySet().forEach(entry ->
+        columns.add(new ServiceNowColumn(entry.getKey(), "string"))
+      );
+      return SchemaBuilder.constructSchema(tableName, columns, legacyMapping);
     }
+    return null;
+  }
+
+  public JsonObject getFirstRecord(JsonReader reader) throws IOException {
+    reader.beginObject();
+    while (reader.hasNext()) {
+      String name = reader.nextName();
+      if (name.equals(ServiceNowConstants.RESULT)) {
+        reader.beginArray();
+        if (reader.hasNext()) {
+          // ONLY parse the first object into memory
+          return GSON.fromJson(reader, JsonObject.class);
+        }
+        reader.endArray();
+      } else {
+        reader.skipValue();
+      }
+    }
+    reader.endObject();
     return null;
   }
 
--- a/src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
+++ b/src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
@@ -16,6 +16,8 @@
 package io.cdap.plugin.servicenow.connector;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.stream.JsonReader;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
@@ -55,6 +57,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -103,7 +108,7 @@ public class ServiceNowConnector implements DirectConnector {
    * Browse Details for the given AccessToken.
    */
   public BrowseDetail browse(ConnectorContext connectorContext,
-                             String accessToken) throws ServiceNowAPIException {
+                             String accessToken) throws ServiceNowAPIException, IOException {
     int count = 0;
     FailureCollector collector = connectorContext.getFailureCollector();
     config.validateCredentialsFields(collector);
@@ -126,7 +131,7 @@ public class ServiceNowConnector implements DirectConnector {
   /**
    * @return the list of tables.
    */
-  private TableList listTables(String accessToken) throws ServiceNowAPIException {
+  private TableList listTables(String accessToken) throws ServiceNowAPIException, IOException {
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       config.getRestApiEndpoint(), OBJECT_TABLE_LIST, false, SchemaType.SCHEMA_API_BASED);
     requestBuilder.setAuthHeader(accessToken);
@@ -135,7 +140,7 @@ public class ServiceNowConnector implements DirectConnector {
     ServiceNowTableAPIClientImpl serviceNowTableAPIClient = new ServiceNowTableAPIClientImpl(config, true);
     RestAPIResponse apiResponse =
         serviceNowTableAPIClient.executeGetWithRetries(requestBuilder.build());
-    return GSON.fromJson(apiResponse.getResponseBody(), TableList.class);
+    return GSON.fromJson(serviceNowTableAPIClient.createJsonReader(apiResponse.getResponseStream()), TableList.class);
   }
 
   public ConnectorSpec generateSpec(ConnectorContext connectorContext, ConnectorSpecRequest connectorSpecRequest) {
@@ -172,7 +177,7 @@ public class ServiceNowConnector implements DirectConnector {
   }
 
   private List<StructuredRecord> getTableData(String tableName, int limit)
-      throws OAuthProblemException, OAuthSystemException, ServiceNowAPIException {
+    throws OAuthProblemException, OAuthSystemException, ServiceNowAPIException, IOException {
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       config.getRestApiEndpoint(), tableName, false, SchemaType.SCHEMA_API_BASED)
       .setExcludeReferenceLink(true)
@@ -183,22 +188,40 @@ public class ServiceNowConnector implements DirectConnector {
     requestBuilder.setAuthHeader(accessToken);
     requestBuilder.setResponseHeaders(ServiceNowConstants.HEADER_NAME_TOTAL_COUNT);
     RestAPIResponse apiResponse = serviceNowTableAPIClient.executeGetWithRetries(requestBuilder.build());
-    List<Map<String, String>> result = serviceNowTableAPIClient.parseResponseToResultListOfMap
-      (apiResponse.getResponseBody());
+    InputStream in = apiResponse.getResponseStream();
+    JsonReader reader = new JsonReader(new InputStreamReader(in, StandardCharsets.UTF_8));
     List<StructuredRecord> recordList = new ArrayList<>();
     Schema schema = getSchema(tableName);
     if (schema != null) {
       List<Schema.Field> tableFields = schema.getFields();
-      for (int i = 0; i < result.size(); i++) {
-        StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
-        for (Schema.Field field : tableFields) {
-          String fieldName = field.getName();
-          ServiceNowRecordConverter.convertToValue(fieldName, field.getSchema(), result.get(i), recordBuilder, false);
+      try {
+        reader.beginObject();
+        while (reader.hasNext()) {
+          String name = reader.nextName();
+          if (name.equals(ServiceNowConstants.RESULT)) {
+            reader.beginArray();
+            while (reader.hasNext()) {
+              JsonObject jsonObject = GSON.fromJson(reader, JsonObject.class);
+              StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
+              for (Schema.Field field : tableFields) {
+                String fieldName = field.getName();
+                ServiceNowRecordConverter.convertToValue(fieldName, field.getSchema(), jsonObject, recordBuilder,
+                  false);
+              }
+              StructuredRecord structuredRecord = recordBuilder.build();
+              recordList.add(structuredRecord);
+            }
+            reader.endArray();
+          } else {
+            reader.skipValue();
+          }
         }
-        StructuredRecord structuredRecord = recordBuilder.build();
-        recordList.add(structuredRecord);
+        reader.endObject();
+      } catch (IOException e) {
+        throw new RuntimeException("Error reading JSON response for table " + tableName, e);
       }
     }
+    apiResponse.close();
     return recordList;
 
   }
--- a/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
+++ b/src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
@@ -21,11 +21,15 @@ import com.google.gson.JsonObject;
 import io.cdap.plugin.servicenow.apiclient.ServiceNowAPIException;
 import io.cdap.plugin.servicenow.util.ServiceNowConstants;
 import org.apache.http.Header;
+import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
-import org.apache.http.util.EntityUtils;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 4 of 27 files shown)


src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
```diff
@@ -152,4 +161,26 @@ public void validateTable(String tableName, SourceValueType valueType, FailureCo
     }
   }
 
+  /**
+   * Check whether the result is empty or not.
```
Sunish-Dahiya: Also explain how are we checking is result is empty

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -197,6 +203,40 @@ public List<Map<String, String>> parseResponseToResultListOfMap(String responseB
     return GSON.fromJson(ja, type);
   }
 
+  public List<Map<String, String>> parseResponseToResultListOfMap(InputStream in) throws ServiceNowAPIException {
```
Sunish-Dahiya: Why are we creating a List of map? Instead we can directly read it as a JsonObject, something like:

gson.fromJson(reader, ServiceNowRecordObject.class);

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -231,12 +271,14 @@ private String getErrorMessage(String responseBody) {
    * @param limit     The number of records to be fetched
    * @return The list of Map; each Map representing a table row
    */
-  public List<Map<String, String>> fetchTableRecordsRetryableMode(String tableName, SourceValueType valueType,
+  public RestAPIResponse fetchTableRecordsRetryableMode(String tableName, SourceValueType valueType,
                                                                   String startDate, String endDate, int offset,
                                                                   int limit) throws ServiceNowAPIException {
-    final List<Map<String, String>> results = new ArrayList<>();
+    //final List<Map<String, String>> results = new ArrayList<>();
```
Sunish-Dahiya: please remove

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -231,12 +271,14 @@ private String getErrorMessage(String responseBody) {
...
+    final RestAPIResponse[] restAPIResponse = new RestAPIResponse[1];
     Callable<Boolean> fetchRecords = () -> {
-      results.addAll(fetchTableRecords(tableName, valueType, startDate, endDate, offset, limit));
+      // results.addAll(fetchTableRecords(tableName, valueType, startDate, endDate, offset, limit));
```
Sunish-Dahiya: please remove

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -231,12 +271,14 @@ private String getErrorMessage(String responseBody) {
...
+      restAPIResponse[0] = fetchTableRecords(tableName, valueType, startDate, endDate, offset, limit);
```
Sunish-Dahiya: can you please explain what are we trying to do here? why restAPIResponse is an array

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -545,8 +592,9 @@ public Map<String, String> getRecordFromServiceNowTable(String tableName, String
    * @throws RuntimeException if the schema response is null or contains no result.
    */
   private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
-                                          String tableName) {
-    List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getResponseBody());
+                                          String tableName) throws ServiceNowAPIException {
+    List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getBodyAsStream());
+    // List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getResponseBody());
```
Sunish-Dahiya: remove

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -545,8 +592,9 @@ public Map<String, String> getRecordFromServiceNowTable(String tableName, String
    * @throws RuntimeException if the schema response is null or contains no result.
    */
   private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
-                                          String tableName) {
-    List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getResponseBody());
+                                          String tableName) throws ServiceNowAPIException {
+    List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getBodyAsStream());
```
Sunish-Dahiya: should be List of Records/objects instead of map

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -82,15 +96,32 @@ public static RestAPIResponse parse(HttpResponse httpResponse, String... headerN
...
+    HttpEntity httpEntity = httpResponse.getEntity();
+    byte[] responseBody = new byte[0];
+    InputStream inputStream;
+    if (httpEntity != null) {
+      inputStream = httpEntity.getContent();
+      BoundedInputStream boundedInputStream = new BoundedInputStream(
+        inputStream, MAX_PAGE_BYTES + 1); // +1 to detect overflow
+      responseBody = IOUtils.toByteArray(boundedInputStream);
+      LOG.info("RAW JSON: {}", new String(responseBody, StandardCharsets.UTF_8));
+      if (responseBody.length > MAX_PAGE_BYTES) {
```
Sunish-Dahiya: what are we trying to do here? why we are loading the complete page in buffer?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -82,15 +96,32 @@ public static RestAPIResponse parse(HttpResponse httpResponse, String... headerN
...
+      ServiceNowAPIException serviceNowAPIException) throws IOException {
+    HttpEntity httpEntity = httpResponse.getEntity();
+    byte[] responseBody = new byte[0];
+    InputStream inputStream;
+    if (httpEntity != null) {
+      inputStream = httpEntity.getContent();
+      BoundedInputStream boundedInputStream = new BoundedInputStream(
+        inputStream, MAX_PAGE_BYTES + 1); // +1 to detect overflow
+      responseBody = IOUtils.toByteArray(boundedInputStream);
+      LOG.info("RAW JSON: {}", new String(responseBody, StandardCharsets.UTF_8));
```
Sunish-Dahiya: Remove as it would unnecessary log the whole response

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
```diff
@@ -82,15 +96,32 @@ public static RestAPIResponse parse(HttpResponse httpResponse, String... headerN
...
-      responseBody = EntityUtils.toString(httpResponse.getEntity());
+      return prepareResponseStream(httpResponse, headers, serviceNowAPIException);
     } catch (IOException e) {
       return new RestAPIResponse(headers, null, new ServiceNowAPIException(e, httpResponse));
     }
-    serviceNowAPIException = validateRestApiResponse(httpResponse, responseBody);
-    return new RestAPIResponse(headers, responseBody, serviceNowAPIException);
+  }
+
+  public static RestAPIResponse prepareResponseStream(HttpResponse httpResponse, Map<String, String> headers,
```
Sunish-Dahiya: didn't understand what are we trying to do with this function? we should be returning a Stream instead of the whole response body

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -121,7 +122,8 @@ public void createPostRequest(Map<String, RestRequest> restRequestsMap, String a
       requestBuilder.setEntity(stringEntity);
       apiResponse = restApi.executePost(requestBuilder.build());
 
-      JsonObject responseJSON = jsonParser.parse(apiResponse.getResponseBody()).getAsJsonObject();
+      JsonObject responseJSON = jsonParser.parse(
+        new InputStreamReader(apiResponse.getBodyAsStream(), StandardCharsets.UTF_8)).getAsJsonObject();
```
Sunish-Dahiya: what are we trying to do here? post requests won't return the records so why we need stream here?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -17,25 +17,42 @@
...
 /**
  * Record reader that reads the entire contents of a ServiceNow table.
  */
 public class ServiceNowMultiRecordReader extends ServiceNowBaseRecordReader {
 
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceNowMultiRecordReader.class);
   private final ServiceNowMultiSourceConfig multiSourcePluginConf;
   private ServiceNowTableAPIClientImpl restApi;
+  private final Gson gson = new Gson();
+  private final Type mapType = new TypeToken<Map<String, String>>() { }.getType();
```
Sunish-Dahiya: should use an object instead of map

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -91,14 +123,87 @@ public StructuredRecord getCurrentValue() throws IOException {
...
+    closeCurrentPage();
+    LOG.debug("Fetching data for table {} at offset {}", tableName, split.getOffset());
+    RestAPIResponse resp = fetchData();
+    LOG.debug("Fetched data for table {} at offset {}", tableName, split.getOffset());
+    InputStream in = resp.getBodyAsStream();
+    if (in == null) {
+      return false;
+    }
+    this.jsonReader = new JsonReader(new InputStreamReader(in, StandardCharsets.UTF_8));
+    this.jsonReader.setLenient(true);
```
Sunish-Dahiya: why we need this?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -91,14 +123,87 @@ public StructuredRecord getCurrentValue() throws IOException {
...
+
+    // Position the reader to the "result" array: { "result": [ ... ], ... }
+    try {
+      JsonToken top;
+      try {
+        top = jsonReader.peek();
+        LOG.info("Peeking JSON token for table {}: {}", tableName, top);
```
Sunish-Dahiya: please remove, it should be a debug log

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
```diff
@@ -91,14 +123,87 @@ public StructuredRecord getCurrentValue() throws IOException {
...
+    }
 
-    iterator = results.iterator();
+    // No "result" array not found, close the current page and return false
+    closeCurrentPage();
+    return false;
+  }
+
+  public void closeCurrentPage() {
+    LOG.info("Closing current page for table {}", tableName);
```
Sunish-Dahiya: not needed?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -103,14 +131,86 @@ public StructuredRecord getCurrentValue() throws IOException {
...
-                                                     pluginConf.getPageSize());
-    LOG.debug("Results size={}", results.size());
+    RestAPIResponse restAPIResponse = restApi.fetchTableRecordsRetryableMode(tableName, pluginConf.getValueType(),
+     pluginConf.getStartDate(), pluginConf.getEndDate(), split.getOffset(), pluginConf.getPageSize());
 
-    iterator = results.iterator();
+    return restAPIResponse;
+  }
+
+  private boolean openNextPage() throws IOException, ServiceNowAPIException {
```
Sunish-Dahiya: looks redundant, as same function is present in src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java

can we abstract these out in a common file. Same comment for nextKeyValue()

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
sgarg-CS: It specifically looks for a top-level key named `result`. Once found, it opens the associated array and checks for the presence of a first element. If it does contain at least one element, then it return `false`, otherwise `true`. 

Updated the method comment. ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Refactored the code to parse the the RestAPIResponse into result. ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

```
APIResponse apiResponse = GSON.fromJson(new JsonReader(new InputStreamReader(in, StandardCharsets.UTF_8)), APIResponse.class);
return apiResponse.getResult();
```

The calling method : `serviceNowTableAPIClient.parseResponseToResultListOfMap` in `ServiceNowConnector.getTableData` expects a result of type: `List<Map<String, String>>`

Can you check the `APIResponse` class once? When you say `ServiceNowRecordObject` class, do you expect that class to hold the result array returned by the ServiceNow Table REST API ? Please clarify.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: removed ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: removed ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: In `Java`, when we use a variable from an outer scope inside a lambda expression or a `Callable`, that variable must be effectively `final`. This means its reference cannot change after it is initialised. so, I was creating an array earlier for `RestAPIResponse` and returning only the first element. I've now replaced it with `AtomicReference`. ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: removed ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: The downstream methods expect `<List<Map<String, String>>`. Are you suggesting to replace the return type with `List<Object>` ? Can you elaborate here?

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: We load one complete page of the ServiceNow API response into a bounded buffer because the plugin requires multiple independent consumers of the response body (validation, processing, error handling, retry decisions), and the HTTP response stream cannot be safely  replayed. An `InputStream` cannot be safely reused across validation, processing, and retry flows due to its single-use nature and HTTP client lifecycle constraints; therefore, the response is materialised into a bounded byte array to ensure deterministic ownership, safe re-exposure, and predictable memory usage.

Storing the response as a bounded byte array is more memory-efficient and predictable than storing it as a String. Strings often require double memory usage, increasing peak heap pressure and GC overhead. The byte[] approach preserves streaming semantics while enforcing strict upper bounds to prevent OOM.

### **Reasoning**
During the streaming implementation, the HTTP response body was obtained as an `InputStream` from the HTTP client. An initial design attempted to store this `InputStream` directly in the `RestAPIResponse` object and reuse it across multiple stages, including validation and downstream processing.

However, this approach led to repeated runtime failures and architectural constraints that could not be resolved without fundamentally changing how the response body is represented.

**Why an InputStream Cannot Be Safely Stored and Reused**

1. InputStream is a single-use, forward-only abstraction
- An InputStream
  - can only be consumed once
  - does not support rewind or replay unless explicitly buffered
  - has no contract guaranteeing reusability across method boundaries

- In our flow, the response body must be accessed by:
  - Validation logic (e.g., empty-result checks)
  - Downstream processing logic
  - Error-handling and retry decision logic
This inherently requires multiple logical consumers, which violates the fundamental design of InputStream.

2. HTTP client–managed streams have strict lifecycle constraints
    The InputStream returned by: `httpResponse.getEntity().getContent()` is owned and managed by the HTTP client.
     
- The stream gets closed when:
  - The enclosing try or resource scope exits
  - The HTTP entity is consumed or garbage-collected
  - A retry or error path is triggered
- As a result, passing the raw stream outside its immediate scope resulted in:
`java.io.IOException` Attempted read on closed stream. This behaviour is expected and by design for HTTP client libraries.

3. Returning an InputStream violates encapsulation and ownership
By embedding an InputStream inside RestAPIResponse:
- Stream ownership becomes unclear
- It is ambiguous which layer is responsible for closing it
- Downstream consumers may accidentally close or exhaust the stream
- Debugging lifecycle errors becomes non-deterministic

This makes the system fragile and highly order-dependent.

4. Validation and processing require independent access

Even when validation is performed before processing, both operations require access to the same response body. Because InputStream consumption is destructive:
- Validation consumes the stream
- Processing later sees an empty or closed stream

This made it impossible to safely separate validation and processing flows using a shared InputStream.

5. Retry logic is incompatible with reused streams

The connector implements retries at the HTTP call level. Reusing an already-consumed `InputStream` across retries is impossible:
- Streams cannot be reset after failure
- Retrying requires either a fresh response or a replayable buffer

Storing the raw stream therefore breaks retry semantics.

**Why Switching to the Byte Array works?**
1. Byte arrays provide deterministic ownership and lifetime
- By eagerly reading the response body into a bounded byte[]:
  - The plugin gains full control over the data lifecycle
  - The HTTP connection can be safely released immediately
  - The response body becomes independent of the HTTP client
This eliminates all stream-closure race conditions.

2. Byte arrays enable safe re-exposure as streams
A `byte[]` allows creation of fresh, independent streams on demand

3. Byte arrays are more memory-efficient than strings
- Compared to storing the response as a String
   - Uses 1 byte per byte of response
   - Avoids temporary decoding buffers
   - Reduces GC pressure
   
This directly lowers the risk of OutOfMemoryErrors.

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Removed ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Refactored the method name from `prepareResponseStream` to `prepareResponse`.

InputStream obtained from the HTTP Response cannot be safely reused and replayed across validation, processing, and retry flows due to its single-use nature and HTTP client lifecycle constraints (InputStream gets closed, when HTTP Connection is closed by try-with-resource block); therefore, the response is materialised into a bounded byte array to ensure deterministic ownership, safe re-exposure, and predictable memory usage.

Detailed reasoning provided, in the earlier comment.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: It is parsing the JSON Response returned by the ServiceNow POST REST API.

Since we've switched to the streaming implementation, we no longer have a method that returns the `responseBody` in the `String` format which the `jsonParser.parse` expected earlier. 
The other alternative is create a `String` from the `byte` array and pass it in the `args`. Will that work for you?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: Do we need to replace `TypeToken<Map<String, String>>()` with` TypeToken<Object>()` here? Please confirm.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: In general, the ServiceNow Table REST API provides standard-compliant JSON that does not strictly require `jsonreader.setlenient(true)` for basic parsing. However, there are specific scenarios where enabling leniency becomes necessary or is a recommended safeguard.

If the API call fails (e.g., 401 Unauthorised, 404 Not Found, or a ServiceNow "maintenance" page), the server might return an HTML body instead of JSON. A strict parser will fail immediately but a lenient one may provide a more readable error or allow you to handle the string manually.

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: Yes, Changed it to `DEBUG` ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiRecordReader.java
sgarg-CS: Yes, had added this logger during dev testing to troubleshoot an issue. Removed it now. ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: Added the implementation of `nextKey()`, `openNextPage()` and `closeCurrentPage()` to the `ServiceNowBaseRecordReader` class ba5eb536e47cd4c00b11fee8766b468e5f0e6dde

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
```diff
@@ -152,4 +161,28 @@ public void validateTable(String tableName, SourceValueType valueType, FailureCo
     }
   }
 
+  /**
+   * Determines if the "result" array in a ServiceNow REST API response is empty. It specifically looks for a top-level
```
Sunish-Dahiya: can you please format it as per java doc.
/** One line summary
*
*  Additional details

* params
*@return
*/

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -231,12 +242,14 @@ private String getErrorMessage(String responseBody) {
    * @param limit     The number of records to be fetched
    * @return The list of Map; each Map representing a table row
    */
-  public List<Map<String, String>> fetchTableRecordsRetryableMode(String tableName, SourceValueType valueType,
+  public RestAPIResponse fetchTableRecordsRetryableMode(String tableName, SourceValueType valueType,
                                                                   String startDate, String endDate, int offset,
                                                                   int limit) throws ServiceNowAPIException {
-    final List<Map<String, String>> results = new ArrayList<>();
+    // Using AtomicReference to capture a value inside a lambda that needs to be accessed outside.
+    AtomicReference<RestAPIResponse> responseRef = new AtomicReference<>();
```
Sunish-Dahiya: this seems like a overkill, it's better to abstract the logic in a function instead of lambda to simplify the code here

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: Instead of a Map, we should use well defined Java Objects.
Since we are already refactoring the code we can always change the caller functions, I don't see that as a blocker

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: I can still see the comment

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -197,6 +202,12 @@ public List<Map<String, String>> parseResponseToResultListOfMap(String responseB
     return GSON.fromJson(ja, type);
   }
 
+  public List<Map<String, String>> parseResponseToResultListOfMap(InputStream in) {
```
Sunish-Dahiya: Should use well defined Objects here instead of Map

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
Sunish-Dahiya: Loading the complete page defeats the whole purpose of this change altogether. We are making this change to avoid OOM issues when we load the entire page into memory and loading it into a bounded buffer is not solving that problem.

Regarding multiple access of the input stream: We can always refactor out code(validation etc.) such that all the processing happens record by record instead of the whole page.
Basically when the connector request for a record we can do all the processing/validation is required, instead of doing it beforehand. This would solve the problem of multiple access of InputStream.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
Sunish-Dahiya: Are we switching to stream for POST methods also? Can't we restrict it to read/GET operations only?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
```diff
@@ -40,11 +54,121 @@ public abstract class ServiceNowBaseRecordReader extends RecordReader<NullWritab
...
   protected Map<String, String> row;
+  protected final Gson gson = new Gson();
+  protected final Type mapType = new TypeToken<Map<String, String>>() { }.getType();
+  protected JsonReader jsonReader = null;
 
   public ServiceNowBaseRecordReader() {
   }
+  
+  /**
+   * The refactored nextKeyValue() — uses Gson JsonReader to stream one record at a time.
```
Sunish-Dahiya: Please rephrase the comment here, refactored doesn't provide any context to the reader, the purpose of comment is not  to provide  the history of the change but explain the logic

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
```diff
@@ -40,11 +54,121 @@ public abstract class ServiceNowBaseRecordReader extends RecordReader<NullWritab
...
+    closeCurrentPage();
+    return false;
+  }
+  
+  public boolean openNextPage() throws IOException, ServiceNowAPIException {
+    LOG.debug("Opening next page for table {} at offset {}", tableName, split.getOffset());
+    closeCurrentPage();
+    LOG.debug("Fetching data for table {} at offset {}", tableName, split.getOffset());
+    RestAPIResponse resp = fetchData();
+    LOG.debug("Fetched data for table {} at offset {}", tableName, split.getOffset());
```
Sunish-Dahiya: do we really need these many debug logs?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
sgarg-CS: Rephrased. PTAL. 611517352dbbb90cbfab170e0f75b530eba32a8a

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Sorry, it's not clear to me what's the ask here. Can you please elaborate more what you mean by well defined objects ? Are you looking for a POJO class here?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Refactored the code by separating out the logic in a private method. PTAL. 611517352dbbb90cbfab170e0f75b530eba32a8a

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
sgarg-CS: Formatted 611517352dbbb90cbfab170e0f75b530eba32a8a

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
sgarg-CS: Not really, these were more for dev-testing. I can remove them. 4b89ca50212af5ef06ddc9bc29f4835ffe43b755

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: If I understand it correctly this Map is used to store the response of the API, if yes then response can be represented by a POJO class instead of a generic map

src/main/java/io/cdap/plugin/servicenow/restapi/RestAPIResponse.java
sgarg-CS: Returning `InputStream` by adding it to the `RestAPIResponse` object instead of a `byte` array didn't work for me. `The Inputstream` gets closed due to the closing of the HTTP Connection. I'm able to build the code but the plugin fails at the runtime while attempting to read data off the stream. 

**Error:** Premature end of chunk coded message body: closing chunk expected.
image


The error can be reproduced by checking out this new [branch](https://github.com/cloudsufi/servicenow-plugins/tree/feature/return_stream) where I'm returning the `InputStream` as a part of RestAPIResponse.

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: Both the GET and POST requests use the same flow for parsing response.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Yes, it is used to store the ServiceNow API Response. Can we use this [POJO class](https://github.com/data-integrations/servicenow-plugins/blob/develop/src/main/java/io/cdap/plugin/servicenow/model/APIResponse.java) ?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Can we use this [APIResponse.java](https://github.com/data-integrations/servicenow-plugins/blob/develop/src/main/java/io/cdap/plugin/servicenow/model/APIResponse.java) as the POJO class here, instead of a Map?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Removed 611517352dbbb90cbfab170e0f75b530eba32a8a

itsankit-google (changes requested):
Please fix the compilation failure in cdap-e2e-tests:

```
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:testCompile (default-testCompile) on project servicenow-plugins: Compilation failure
[ERROR] /tmp/servicenow-plugins/servicenow-plugins/plugin/src/e2e-test/java/io/cdap/plugin/servicenowsink/actions/ServiceNowSinkPropertiesPageActions.java:[66,78] incompatible types: com.google.gson.JsonObject cannot be converted to java.util.Map<java.lang.String,java.lang.String>
```

itsankit-google (changes requested):
Still failures with JsonSyntaxExcsption : https://storage.googleapis.com/e2e-tests-cucumber-reports/servicenow-plugins/refs/pull/131/merge/cucumber-reports/advanced-reports/cucumber-html-reports/report-feature_4_2095121949.html

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -177,6 +183,12 @@ public List<Map<String, String>> parseResponseToResultListOfMap(String responseB
     return GSON.fromJson(ja, type);
   }
 
+  public List<JsonObject> parseResponseToResultListOfMap(InputStream in) {
```
itsankit-google: this method is returning `List<JsonObject>` but the name of method is `parseResponseToResultListOfMap`, please fix.

pom.xml
```diff
@@ -99,6 +99,11 @@
       <version>${cdap.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.5</version>
```
itsankit-google: why using `2.5` when newer versions are available?

pom.xml
```diff
@@ -99,6 +99,11 @@
       <version>${cdap.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
```
itsankit-google: Also please move the version to properties section.

pom.xml
sgarg-CS: This is unused now. Removed it. 016cdfb601d2ec6c0fe7d8b8f5fd6e1a3ee6a9d7

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Renamed the method accordingly. 016cdfb601d2ec6c0fe7d8b8f5fd6e1a3ee6a9d7

pom.xml
sgarg-CS: Not needed now.

itsankit-google (changes requested):
keeping this comment until the e2e test is fixed : 

Image

sgarg-CS:
> keeping this comment until the e2e test is fixed :
> 
> Image

This e2e-test failure has been fixed. d229af47d93013dbfb125b3301443014b8dda785
[PR](https://github.com/cdapio/cdap-e2e-tests/pull/298) has been raised to fix another e2e-test failure related to the window size.
`org.openqa.selenium.TimeoutException: Expected condition failed: waiting for element to be clickable: [[ChromeDriver: chrome on linux (0dccaeaec775cf076e469f7a5fdf64f6)] -> xpath: //div[@data-cy='pipeline-metadata']] (tried for 300 second(s) with 500 milliseconds interval)
`

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -177,6 +182,12 @@ public List<Map<String, String>> parseResponseToResultListOfMap(String responseB
     return GSON.fromJson(ja, type);
   }
 
+  public List<JsonObject> parseResponseToResultListOfJsonObject(InputStream in) {
```
Sunish-Dahiya: name is very verbose, better to rename it to something like `getApiResponse`/
parseApiResponse

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: Why not return the response from the function instead of capturing it inside a variable. Using reference makes it very hard to read

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -531,13 +556,12 @@ public Map<String, String> getRecordFromServiceNowTable(String tableName, String
    * @throws RuntimeException if the schema response is null or contains no result.
    */
   private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
-                                          String tableName) {
-    List<Map<String, String>> result = parseResponseToResultListOfMap(restAPIResponse.getResponseBody());
+                                          String tableName) throws ServiceNowAPIException {
+    List<JsonObject> result = parseResponseToResultListOfJsonObject(restAPIResponse.getResponseStream());
```
Sunish-Dahiya: This is again loading the whole response object in memory right?
We only need the first record for schema right(as mentioned in comment)?

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
```diff
@@ -121,7 +123,8 @@ public void createPostRequest(Map<String, RestRequest> restRequestsMap, String a
       requestBuilder.setEntity(stringEntity);
       apiResponse = restApi.executePost(requestBuilder.build());
 
-      JsonObject responseJSON = jsonParser.parse(apiResponse.getResponseBody()).getAsJsonObject();
+      JsonObject responseJSON = jsonParser.parse(new JsonReader(new InputStreamReader(apiResponse.getResponseStream())))
```
Sunish-Dahiya: we can abstract the reader code in a function: createJsonReader and use that everywhere.

Also why getAsJsonObject() is getting called twice?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
```diff
@@ -39,12 +54,132 @@ public abstract class ServiceNowBaseRecordReader extends RecordReader<NullWritab
   protected String tableNameField;
   protected List<Map<String, String>> results;
   protected Iterator<Map<String, String>> iterator;
-  protected Map<String, String> row;
+  protected JsonObject row;
+  protected final Gson gson = new Gson();
+  protected final Type mapType = new TypeToken<Map<String, String>>() { }.getType();
```
Sunish-Dahiya: where is mapType getting used?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
```diff
@@ -39,12 +54,132 @@ public abstract class ServiceNowBaseRecordReader extends RecordReader<NullWritab
   protected String tableNameField;
   protected List<Map<String, String>> results;
   protected Iterator<Map<String, String>> iterator;
```
Sunish-Dahiya: Where are we using these variables?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
```diff
@@ -183,8 +184,8 @@ private List<StructuredRecord> getTableData(String tableName, int limit)
     requestBuilder.setAuthHeader(accessToken);
     requestBuilder.setResponseHeaders(ServiceNowConstants.HEADER_NAME_TOTAL_COUNT);
     RestAPIResponse apiResponse = serviceNowTableAPIClient.executeGetWithRetries(requestBuilder.build());
-    List<Map<String, String>> result = serviceNowTableAPIClient.parseResponseToResultListOfMap
-      (apiResponse.getResponseBody());
+    List<JsonObject> result = serviceNowTableAPIClient.parseResponseToResultListOfJsonObject
```
Sunish-Dahiya: same as above why we are loading the whole response in memory here?

Sunish-Dahiya (changes requested):

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: We already have similar named method in the RestAPIResponse class. Can we rename it to `parseInputStreamToJsonList` ?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
sgarg-CS: Refactored to code for the CDAP specific sampling functionality to read one record at a time. 12335802f565f20bcc16a2d4186bfbfdf412ac52

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
sgarg-CS: Unused now. Removed them. 12335802f565f20bcc16a2d4186bfbfdf412ac52

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
sgarg-CS: Unused now. Removed it. 12335802f565f20bcc16a2d4186bfbfdf412ac52

src/main/java/io/cdap/plugin/servicenow/sink/service/ServiceNowSinkAPIRequestImpl.java
sgarg-CS: > we can abstract the reader code in a function: createJsonReader and use that everywhere.

Created a separate function : `getJsonReader` to hold the `JsonReader` code. 12335802f565f20bcc16a2d4186bfbfdf412ac52

> Also why getAsJsonObject() is getting called twice?

Are you referring this block of code?

```
int statusCode = servicedRequestsArray.get(i).getAsJsonObject().get(ServiceNowConstants.STATUS_CODE)
                .getAsInt();
        if (statusCode / 100 == 4 || statusCode / 100 == 5) {
          String encodedResponseBody = servicedRequestsArray.get(i).getAsJsonObject().get(ServiceNowConstants.BODY)
                  .getAsString();
          String decodedResponseBody = new String(Base64.getDecoder().decode(encodedResponseBody));
          String errorDetail = jsonParser.parse(decodedResponseBody).getAsJsonObject().get(ServiceNowConstants.ERROR)
                  .getAsJsonObject().get(ServiceNowConstants.ERROR_DETAIL).getAsString();
```
                  
This is the logic for handling of the response from ServiceNow Batch Insertion API

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Yes, but this was only for schema construction. I've refactored the code, to only parse the first object into memory. 12335802f565f20bcc16a2d4186bfbfdf412ac52

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Changed the return type from `Boolean` to `RestAPIResponse` for the Retryer and returning the response directly instead of storing it in a separate variable. 12335802f565f20bcc16a2d4186bfbfdf412ac52

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
```diff
@@ -183,22 +188,40 @@ private List<StructuredRecord> getTableData(String tableName, int limit)
...
     if (schema != null) {
       List<Schema.Field> tableFields = schema.getFields();
-      for (int i = 0; i < result.size(); i++) {
-        StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
-        for (Schema.Field field : tableFields) {
-          String fieldName = field.getName();
-          ServiceNowRecordConverter.convertToValue(fieldName, field.getSchema(), result.get(i), recordBuilder, false);
+      try {
+        reader.beginObject();
+        while (reader.hasNext()) {
```
Sunish-Dahiya, line 199: can't we use the record reader here instead of implementing it again?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -530,19 +542,45 @@ public Map<String, String> getRecordFromServiceNowTable(String tableName, String
...
+      if (firstRecord != null) {
+      firstRecord.entrySet().forEach(entry ->
+        columns.add(new ServiceNowColumn(entry.getKey(), "string"))
+      );
       return SchemaBuilder.constructSchema(tableName, columns, false);
     }
     return null;
   }
 
+  private JsonObject getFirstRecord(JsonReader reader) throws IOException {
```
Sunish-Dahiya: can we reuse record reader?

src/main/java/io/cdap/plugin/servicenow/ServiceNowBaseConfig.java
```diff
@@ -83,7 +92,7 @@ public void validateCredentials(FailureCollector collector) {
   }
 
   @VisibleForTesting
-  public void validateServiceNowConnection(FailureCollector collector) {
+  public void   validateServiceNowConnection(FailureCollector collector) {
```
Sunish-Dahiya: empty space, please fix formatting

sgarg-CS: Removed extra space a5b9371c7de25fecd5fe3e48d786a00f00485978

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Not completely, there are only a handful of code lines mentioned below that are reusable 

```
jsonReader.beginObject();
        while (jsonReader.hasNext()) {
          String name = jsonReader.nextName();
          if (name.equals(ServiceNowConstants.RESULT)) {
            if (jsonReader.peek() == JsonToken.BEGIN_ARRAY) {
              jsonReader.beginArray();
```

Having a separate utility class to keep the logic of reading the object partially and rest at the the source doesn't seem to offer any real advantage, IMO.

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
sgarg-CS: No, the `ServiceNowBaseRecordReader.openNextPage()` cannot be reused as it is. It can only used partially i.e. the initial reading of the object and array traversal.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
Sunish-Dahiya: parseInputStreamToJsonList is better than the current one

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
```diff
@@ -16,35 +16,163 @@
...
 
 /**
  * Base Record reader class that provides a basic structure for Derived Record Reader classes.
  */
 public abstract class ServiceNowBaseRecordReader extends RecordReader<NullWritable, StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceNowRecordReader.class);
   protected ServiceNowInputSplit split;
   protected int pos;
   protected List<Schema.Field> tableFields;
   protected Schema schema;
```
Sunish-Dahiya, line 47: where is this getting used? if in child class then it should be declared in child class. Also can you please check for all the variable and removed unused ones

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
```diff
@@ -105,14 +85,12 @@ public StructuredRecord getCurrentValue() throws IOException {
     return recordBuilder.build();
   }
 
-  private void fetchData() throws ServiceNowAPIException {
+  RestAPIResponse fetchData() throws ServiceNowAPIException {
```
Sunish-Dahiya: What's reason behind removing private? will we be using it somewhere else

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
Sunish-Dahiya: We can create a instance of RecordReader here with split size equal to sample size required. Then we can just read records from the recordreader as we do in connectors code, do you see any blocker in doing that?

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseRecordReader.java
sgarg-CS: There is no change made to this piece of code. Can we please limit the scope for review on this PR to the changes made for streaming implementation only and not the (unchanged) existing code? I've created an improvement story on the CDAP PLUGIN Board [PLUGIN-1946](https://cdap.atlassian.net/browse/PLUGIN-1946). This can be prioritised after discussion with the CDF & DTS S&M team PM and TL's.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Done 3d331c22f0770fa706d2214185f45f29df800d0e

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowRecordReader.java
sgarg-CS: This is one of the overridden implementations of `fetchData()` which is declared as abstract in in the super class: `ServiceNowBaseRecordReader.java` and as per Java Rules, we cannot declare an abstract method private.

Sunish-Dahiya: Then we should add Override annotation

sgarg-CS: Added 35096142cb4721a136e9334da9f427e9b8937df7

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
sgarg-CS: The `ServiceNowRecordReader` requires an instance of `ServiceNowSourceConfig` class in it's constructor args. The `ServiceNowConnector` class where you want to create the `RecordReader` instance doesn't have the `ServiceNowSourceConfig` object. It only has access to `ServiceNowConnectorConfig`. 

Also, this `ServiceNowConnector` class is part of another plugin : `Connector.PLUGIN_TYPE` and not `BatchSource.PLUGIN_TYPE` which is what is used by the DTS ServiceNow Connector

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -177,6 +181,12 @@ public List<Map<String, String>> parseResponseToResultListOfMap(String responseB
     return GSON.fromJson(ja, type);
   }
 
+  public List<JsonObject> parseInputStreamToJsonList(InputStream in) {
```
Sunish-Dahiya: where is this getting used, i didn't find any reference?
Also this function is loading the whole response in memory which is defeating the whole purpose of this PR?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -530,19 +542,45 @@ public Map<String, String> getRecordFromServiceNowTable(String tableName, String
    *
    * @throws RuntimeException if the schema response is null or contains no result.
    */
-  private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
+    private Schema prepareStringBasedSchema(RestAPIResponse restAPIResponse, List<ServiceNowColumn> columns,
```
Sunish-Dahiya, line 543: for my understanding: When we already fetching schema from a metadata API so what is the use case of this?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
Sunish-Dahiya: Then can we abstract this iteration logic in a separate PageIterator class? so that we can avoid duplicate code. Currently the same logic is present in RecordReader, getFirstRecord and this function.

In current state if there is any change in how we parse results then it will require modifications in multiple files which is not the right state to be in.

If you want to cover this in separate PR, but please create a tracking work item for that

Sunish-Dahiya (commented):
Again reiterating the primary purpose of this PR: We shouldn't load the whole data response page into memory at any stage. Loading schema or metadata response is fine as size of that would be within limits only.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: Yes, this method is unused now after refactoring done in the last commits. Removed it and it's reference from the test classes. 84623af37ea2cdc11f393c8dfe7566cb15c8c8a3

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
sgarg-CS: This is to support users who're still using the previous release version : `v1.1`  
Please refer to this comment in the code.
```
// use connection was added in release/1.2, so it will be null for users who are upgrading from release/1.1
    // This is added to support backward compatibility for 1.1 users.
    if (useConnection == null) {
      return SchemaType.STRING_BASED;
    }
```

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnector.java
sgarg-CS: Created a ticket [PLUGIN-1947](https://cdap.atlassian.net/browse/PLUGIN-1947) to track this item on the PLUGIN JIRA Board.

Sunish-Dahiya (approved):
Can you attach the testing reports:
1. We should test for all the existing scenarios.
2. We should also test for high row sizes to ensure OOM is not happening now.

Please do this before merging the change.

PR #134, opened 2026-02-03
Adding filter Predicate support for servicenow

AbhishekKumar9984:


src/main/java/io/cdap/plugin/servicenow/source/ServiceNowBaseSourceConfig.java
```diff
@@ -80,6 +80,12 @@ public class ServiceNowBaseSourceConfig extends ServiceNowBaseConfig {
   @Description("Enable support for new data types such as array (glide_list) ")
   private Boolean enableNewDataTypes;
 
+  @Name(ServiceNowConstants.PROPERTY_FILTER_QUERY)
```
prince-cs: This is supposed to come from DTS side inside of split. Keep start date and end date. From dts side, the split will have filter query you can append that and on CDF side we can keep the same flow of query creation with start and end date

PR #136, opened 2026-02-04
[PLUGIN-1944] Support for Filter Queries in Split class.

AbhishekKumar9984:
Support DTS Filter Queries in Split class.

Overview: To ensure consistency between CDF and DTS, this change updates the Split class to handle filter queries directly. While CDF currently supports filtering via start/end dates, this PR streamlines that process into a more flexible filterQuery format.

Added filterQuery as a new parameter in the Split class.

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
@@ -122,24 +122,26 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   }
 
   /**
-   * Fetch the list of records from ServiceNow table.
+   * Fetches a list of records from a ServiceNow table.
    *
-   * @param tableName The ServiceNow table name
-   * @param valueType The value type
-   * @param startDate The start date
-   * @param endDate   The end date
-   * @param offset    The number of records to skip
-   * @param limit     The number of records to be fetched
-   * @return The list of Map; each Map representing a table row
+   * @param tableName The name of the ServiceNow table.
+   * @param valueType The type of value to retrieve (e.g., actual or display).
+   * @param filterQuery An encoded query to filter the records. For example, to filter records created
+   *  between two dates, the query would be: {@code
+   *  sys_created_onBETWEENjavascript:gs.dateGenerate
+   *  ('2023-01-01','start')@javascript:gs.dateGenerate('2023-01-31','end')}
+   * @param offset The number of records to skip from the beginning of the result set.
+   * @param limit The maximum number of records to retrieve.
+   * @return A list of maps, where each map represents a record from the table.
+   * @throws ServiceNowAPIException If an error occurs while fetching the records.
    */
   public List<Map<String, String>> fetchTableRecords(
-      String tableName,
-      SourceValueType valueType,
-      String startDate,
-      String endDate,
-      int offset,
-      int limit)
-      throws ServiceNowAPIException {
+    String tableName,
+    SourceValueType valueType,
+    String filterQuery,
+    int offset,
+    int limit)
+    throws ServiceNowAPIException {
     ServiceNowTableAPIRequestBuilder requestBuilder = new ServiceNowTableAPIRequestBuilder(
       this.conf.getRestApiEndpoint(), tableName, false, schemaType)
       .setExcludeReferenceLink(true)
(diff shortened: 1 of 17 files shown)


src/main/java/io/cdap/plugin/servicenow/source/ServiceNowMultiInputFormat.java
```diff
@@ -127,6 +128,11 @@ public static Set<String> getList(String value) {
   @Override
   public List<InputSplit> getSplits(JobContext jobContext) throws IOException, InterruptedException {
     ServiceNowJobConfiguration jobConfig = new ServiceNowJobConfiguration(jobContext.getConfiguration());
+
+      String startdate = jobConfig.getPluginConf().getStartDate();
```
vikasrathee-cs: this is not the plugin config for multi. it is multipluginconfig.

src/main/java/io/cdap/plugin/servicenow/util/Util.java
```diff
@@ -56,4 +64,27 @@ public static boolean isValidDateFormat(String dateFormat, String value) {
...
+  public static String generateDateRangeQuery(String startDate, String endDate) {
+    if (Util.isNullOrEmpty(startDate) || Util.isNullOrEmpty(endDate)) {
+      return "";
+    }
+
+    String dateRange = "";
+    try {
+      String createdOnDateRange = String.format(DATE_RANGE_TEMPLATE, FIELD_CREATED_ON, startDate, endDate);
+      String updatedOnDateRange = String.format(DATE_RANGE_TEMPLATE, FIELD_UPDATED_ON, startDate, endDate);
+      dateRange = String.format("%s^OR%s", createdOnDateRange, updatedOnDateRange);
```
vikasrathee-cs, line 84: add a logger to print query

src/main/java/io/cdap/plugin/servicenow/source/ServiceNowInputFormat.java
```diff
@@ -16,6 +16,8 @@
 
 package io.cdap.plugin.servicenow.source;
 
+import com.google.common.base.Strings;
+import io.cdap.cdap.api.PlatformInfo;
```
vikasrathee-cs: what is these new  imports for?

AbhishekKumar9984: unused so removed .

AbhishekKumar9984 (commented):
e2e test failures have been observed due to underlying CDF framework issues , I had tested the logic manually and  also validated it . I have successfully verified these changes by running a CDAP pipeline specifically targeting the previously failing scenarios to ensure the filter conversion and data retrieval work as expected.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -126,20 +126,17 @@ public String getAccessTokenRetryableMode() throws ExecutionException, RetryExce
    *
    * @param tableName The ServiceNow table name
    * @param valueType The value type
-   * @param startDate The start date
-   * @param endDate   The end date
    * @param offset    The number of records to skip
```
Sunish-Dahiya: fix the javadoc, also document the format of this filterQuery

Sunish-Dahiya (commented):
Didn't understand the purpose of this change. Can you please add the details in the description and also explain why this is needed

vikasrathee-cs:
> Didn't understand the purpose of this change. Can you please add the details in the description and also explain why this is needed

Currently CDF supports creation of filter using start date and end date which is slight different than what we need for DTS.  So, added a generic approach to solve it in split class.  Added the details in the description.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -122,24 +122,26 @@ public String getAccessTokenRetryableMode() throws ExecutionException, RetryExce
...
-   * @param startDate The start date
-   * @param endDate   The end date
-   * @param offset    The number of records to skip
-   * @param limit     The number of records to be fetched
-   * @return The list of Map; each Map representing a table row
+   * @param tableName The name of the ServiceNow table.
+   * @param valueType The type of value to retrieve (e.g., actual or display).
+   * @param filterQuery An encoded query to filter the records. For example, to filter records created
+   *  between two dates, the query would be: {@code
+   *  sys_created_onBETWEENjavascript:gs.dateGenerate
```
Sunish-Dahiya, line 131: It's better to add the supported format, may be a link docs will help

Sunish-Dahiya (approved):
Can you please add some testing details in the description.
Also check e2e test as they are failing for this PR.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -434,13 +412,17 @@ public int getTableRecordCount(String tableName)
    * @return the table record count
    * @throws ServiceNowAPIException
    */
-  public int getTableRecordCount(String tableName, String accessToken) throws ServiceNowAPIException {
+  public int getTableRecordCount(String tableName, String accessToken,
```
vikasrathee-cs: this method just needs filterquery not start and end date and make it @nullable

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -421,9 +398,10 @@ private Schema prepareSchemaWithMetadataAPI(RestAPIResponse restAPIResponse, Lis
    * @return the table record count
    * @throws ServiceNowAPIException
    */
-  public int getTableRecordCount(String tableName)
+  public int getTableRecordCount(String tableName, String startdate, String enddate)
```
vikasrathee-cs: Use @Nullable in every method as these can be null
e.g.  @Nullable String startdate, @Nullable String enddate

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
vikasrathee-cs: I can still see start date and end date here

AbhishekKumar9984: adding those in this function so that we can use this at line no. 401

vikasrathee-cs (approved):

PR #139, opened 2026-02-06
PLUGIN-1943: Fix for array support when fetching display values

sgarg-CS:
[PLUGIN-1943](https://cdap.atlassian.net/browse/PLUGIN-1943)

**Issue:**
This PR fixes an issue where `glide_static_list` field types were not mapped to ARRAY<STRING> when the `legacyMapping` flag was set to false

Previously, the ARRAY handling logic only applied to fields of type `glide_list` and actual values. However, when `sysparm_display_value=true` is used, ServiceNow exposes these fields as `glide_static_list`. As a result, display value ingestion incorrectly mapped these fields as `STRING` even with the `legacyMapping`  flag set to false, causing schema inconsistencies.

**Fix:**

Extended schema construction logic to treat both `glide_list` and `glide_static_list` as eligible for `ARRAY<STRING>` mapping when `legacyMapping` flag is set to `false`. Also added unit test coverage for display value mode.

[PLUGIN-1943]: https://cdap.atlassian.net/browse/PLUGIN-1943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
+++ b/src/main/java/io/cdap/plugin/servicenow/util/SchemaBuilder.java
@@ -79,6 +79,7 @@ public class SchemaBuilder {
       case "glide_time":
         return Schema.of(Schema.LogicalType.TIME_MICROS);
       case "glide_list":
+      case "glide_static_list":
         return legacyMapping.equals(Boolean.FALSE) ? Schema.arrayOf(Schema.of(Schema.Type.STRING)) :
           Schema.of(Schema.Type.STRING);
       case "reference":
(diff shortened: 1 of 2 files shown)


vikasrathee-cs (approved):
LGTM

itsankit-google (approved):

itsankit-google:
Curious question, is this case only with `glide_list` type when `sysparm_display_value=true` or with other types as well?

If yes, Why I don't see similar handling for others?

sgarg-CS:
> Curious question, is this case only with `glide_list type` when `sysparm_display_value=true` or with other types as well?
If yes, Why I don't see similar handling for others?

Saw the same behaviour for fields with `choice` type as well, but their internal type: `string` or `integer` are already handled. For rest of the data types, the field `type` and `internal_type` values are same.

PR #151, opened 2026-06-27
[PLUGIN-1959] Add proxy support for ServiceNow plugins

vikasrathee-cs:
Route all ServiceNow API calls (table reads, schema fetch, batch sink writes, OAuth token requests) through an optional HTTP proxy, modeled on the http-plugins proxy support.

- Add proxyUrl, proxyUsername, proxyPassword properties to ServiceNowConnectorConfig (nullable, macro-enabled) with validation
- Build a proxy-aware Apache HttpClient in RestAPIClient used by all GET/POST/DELETE calls
- Route the OAuth token call through the proxy via a new ProxyOltuHttpClient (Oltu HttpClient backed by Apache HttpClient); falls back to URLConnectionClient when no proxy is configured
- Add Proxy fields and a "Proxy authentication" filter to the connector, source, multi-source and sink widgets
- Document the new properties and add unit tests for proxy config

- ProxyOltuHttpClientTest: covers the OAuth-over-proxy client (POST token parsing, GET path, IOException wrapped as OAuthSystemException)
- RestAPIClientProxyTest: covers getHttpClientBuilder() proxy wiring (proxy host + credentials, proxy-only, no-proxy, invalid URL) by inspecting the real HttpClientBuilder's fields, since its setters are final.

Failed e2e are actully getting successful https://storage.googleapis.com/e2e-tests-cucumber-reports/servicenow-plugins/refs/pull/151/merge/cucumber-reports/advanced-reports/cucumber-html-reports/report-feature_8_566810399.html

And one e2e failure is due to slow response from service now instance and canCommit error is coming intermittently.

### DIFF
--- a/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
+++ b/src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
@@ -49,7 +49,6 @@ import org.apache.http.HttpEntity;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpDelete;
 import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClients;
 import org.apache.oltu.oauth2.common.exception.OAuthProblemException;
 import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
 import org.slf4j.Logger;
@@ -86,6 +85,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   public static JsonArray serviceNowJsonResultArray;
 
   public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf, Boolean useConnection) {
+    super(conf.getProxyUrl(), conf.getProxyUsername(), conf.getProxyPassword());
     this.conf = conf;
     this.schemaType = getSchemaTypeBasedOnUseConnection(useConnection);
   }
@@ -561,7 +561,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
     deleteRequest.setHeader("Authorization", "Bearer " + accessToken);
     deleteRequest.setHeader("Accept", "application/json");
 
-    try (CloseableHttpClient httpClient = HttpClients.createDefault();
+    try (CloseableHttpClient httpClient = getHttpClientBuilder().build();
          CloseableHttpResponse response = httpClient.execute(deleteRequest)) {
 
       int statusCode = response.getStatusLine().getStatusCode();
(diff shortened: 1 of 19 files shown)


src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
```diff
@@ -120,6 +171,40 @@ public void validateCredentialsFields(FailureCollector collector) {
...
+  }
+
+  /**
+   * Validates the optional proxy fields. The proxy URL must be a valid host with a protocol and port, and a proxy
+   * username and password must be provided together.
+   */
+  public void validateProxyFields(FailureCollector collector) {
+    if (!containsMacro(ServiceNowConstants.PROPERTY_PROXY_URL) && !Util.isNullOrEmpty(proxyUrl)) {
+      try {
+        HttpHost.create(proxyUrl);
```
shruthi713: As mentioned below by gemini-code-assist, the HttpHost.create() method in the Apache HttpComponents library is intentionally lenient and does accept URLs without an http:// or https:// prefix.so we might need to add an explicit check for those?

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
```diff
@@ -86,6 +85,7 @@ public class ServiceNowTableAPIClientImpl extends RestAPIClient {
   public static JsonArray serviceNowJsonResultArray;
 
   public ServiceNowTableAPIClientImpl(ServiceNowConnectorConfig conf, Boolean useConnection) {
+    super(conf.getProxyUrl(), conf.getProxyUsername(), conf.getProxyPassword());
```
shruthi713, line 88: can we check below suggestion ? is it better to use  `conf.hasProxyConfigured() ? conf.getProxyUrl():null` ?

src/main/java/io/cdap/plugin/servicenow/connector/ServiceNowConnectorConfig.java
vishwasvaidya-cloudsufi: I verified this behavior. HttpHost.create() intentionally defaults to http when no scheme is provided. Based on the discussion, that's acceptable since users are expected to specify https:// explicitly when needed.

src/main/java/io/cdap/plugin/servicenow/apiclient/ServiceNowTableAPIClientImpl.java
vishwasvaidya-cloudsufi: - I tested the proxy macro (e.g., ${purl}) in CDAP, and it actually validated and executed successfully without this change. 
- The crash the bot mentioned doesn't happen because CDAP automatically skips connection validation in the UI when it detects a macro. 
- At runtime, the macro is fully resolved before the HTTP client is built, so no errors are thrown. Since the CDAP framework handles it , the change isn't necessary.

shruthi713 (approved):

